PR #101, opened 2021-01-23
Add angular examples

DanWahlin:
This adds Angular examples for Brainstorm, Collaboration Text, and DiceRoller.


angular/brainstorm/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "fluid-angular",
```
micahgodbolt, line 2: duplicate package names

@DanWahlin can you make the 3 packages have unique names and add them to the learna.json file?

angular/brainstorm/angular.json
```diff
@@ -0,0 +1,128 @@
...
+        }
+      },
+      "root": "",
+      "sourceRoot": "src",
+      "prefix": "app",
+      "architect": {
+        "build": {
+          "builder": "@angular-devkit/build-angular:browser",
+          "options": {
+            "outputPath": "dist/FluidAngular",
```
micahgodbolt, line 20: https://angular.io/guide/build#configuring-commonjs-dependencies
warnings for debug, assert, @fluidframework/container-runtime-definitions and others. 

image

angular/brainstorm/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "dependencies": {
+    "@angular/animations": "~10.1.2",
+    "@angular/common": "~10.1.2",
+    "@angular/compiler": "~10.1.2",
+    "@angular/core": "~10.1.2",
+    "@angular/forms": "~10.1.2",
+    "@angular/platform-browser": "~10.1.2",
+    "@angular/platform-browser-dynamic": "~10.1.2",
+    "@angular/router": "~10.1.2",
+    "@fluidframework/aqueduct": "^0.32.3",
```
micahgodbolt, line 24: bump to 33.3 for all of the fluid packages

angular/brainstorm/angular.json
DanWahlin: Those are normal. Angular warns when CommonJS is used due to tree shaking issues.

angular/brainstorm/package.json
DanWahlin: Sounds like this will be automated so going to leave as is unless it needs to be bumped first before the automation you mention in the Teams chat kicks in. Let me know.

angular/brainstorm/package.json
DanWahlin: Have these updated to be unique in package.json. Haven't used Lerna so wasn't sure about updating that part (path based I'd assume?).

PR #193, opened 2021-05-25
Update brainstorm to the Let's Brainstorm Application

SamBroner:


brainstorm/README.md
```diff
@@ -1,99 +1,30 @@
...
-
-Once you run `start` you can navigate to `http://localhost:8080` in any browser window to use your fluid example.
-
-> The Tinylicious Fluid server must be running. See [`start:server`](###-start:server) below.
-
-### `start:server`
-
-```bash
-npm run start:server
+This example is using an experimental API surface, please be cautious. We will break it!
```
tylerbutler, line 4:
```suggestion
This example is using an experimental API surface so please be cautious. We will break it!
```

tylerbutler (approved):

PR #208, opened 2021-06-15
fix port number

danielroney:
changed tinylicious port to 7070


SamBroner (approved):
This looks great. Thanks for making this change @danielroney

PR #240, opened 2021-08-05
Update brainstorm example to current best practices

sumedhb1995:
- Port over changes made in fluid-dev-experiences to use FrsClient to fetch the fluidContainer and use audience information from containerServices
- Update dependencies to FF 0.44
- Add rendering of liked user lists and members that was missing in this version of the example
- Use fluid-framework package to bring in runtime dependencies in unified import
- Fix usage of useEffect to correctly take off event listeners


brainstorm/src/BrainstormModel.ts
```diff
@@ -127,5 +127,9 @@ export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
     setChangeListener(listener: () => void): void {
       sharedMap.on("valueChanged", listener);
     },
+
+    removeChangeListener(listener: () => void): void {
+      sharedMap.off("valueChanged", listener);
+    }
```
timtwang, line 133: what's the purpose of this method? Seems like this method does not exist in `LetsBrainstormHelloWorld`

brainstorm/src/BrainstormModel.ts
```diff
@@ -65,15 +65,15 @@ export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
...
         color: sharedMap.get(c_ColorPrefix + noteId),
       };
       return newNote;
     },
     
-    GetNoteLikedUsers(noteId: string): IUser[] {
+    GetNoteLikedUsers(noteId: string): FrsMember[] {
       return (
         Array.from(sharedMap
           .keys())
```
timtwang, line 79: there are some comments here in `LetsBrainstormHelloWorld` and I think adding them here will help with future reference.

sumedhb1995: Good call, added!

brainstorm/src/BrainstormModel.ts
sumedhb1995: Yep, I added it in to fix this warning that we get on the creation flow
image
This happens because in `NoteSpace`, we use a useEffect hook to set an event listener that responds to changes in Fluid data, but we were never taking it off once the element got dismounted. The return value of a useEffect hook allows you to set up logic to cleanup any subscriptions like this: https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup
So now, when the hook returns we call `removeChangeListener` to clear that listener

brainstorm/src/view/Header.tsx
```diff
@@ -87,11 +89,20 @@ export function Header(props: HeaderProps) {
     },
   ];
 
-
+  const farItems: ICommandBarItemProps[] = [
+    {
+      key: "presence",
+      onRender: () => <Facepile
```
timtwang, line 95: Just want to confirm, we removed the function to get user photo for personas since that's only available through graph client?

brainstorm/src/view/PersonaList.tsx
```diff
@@ -1,34 +1,27 @@
...
   Persona,
   PersonaSize,
 } from "@fluentui/react";
+import { FrsMember } from "@fluid-experimental/frs-client";
 import React from "react";
-import { IUser } from "../utils";
 
 export type PersonaListProps = {
-  users: IUser[];
+  users: FrsMember[];
```
timtwang, line 11: is it necessary to have a type with just 1 property? I think we can just remove the type altogether and just pass in `FrsMember[]` directly, similar to what we are doing with `BrainstormView`.

brainstorm/src/view/Header.tsx
sumedhb1995: Correct, there's no profile image to fetch. Instead, Persona by default just renders the first character of the username as the profile picture

brainstorm/src/BrainstormModel.ts
timtwang: I see, that make sense. 😃

brainstorm/src/view/PersonaList.tsx
sumedhb1995: Updated, I also cleaned up some other types so that we could drop most of the usages of ts-ignore

timtwang (approved):

PR #247, opened 2021-08-09
Move CRA demo

sumedhb1995:


micahgodbolt:
is this going to be followed up by a removal of cra demo from FluidFramework?

sumedhb1995:
> is this going to be followed up by a removal of cra demo from FluidFramework?

Yep, we can remove it from FF once we have it checked in here to avoid having the same code in two places

micahgodbolt (approved):

PR #248, opened 2021-08-09
Move fluid react starter and update dependencies to latest

sumedhb1995:
- Move fluid-react-starter over
- Update to latest FF dependencies
- Use uploaded image blobs


micahgodbolt:
would like to see the png files pulled out of the repo. They can just as easily be uploaded via adding them to an issue, and then pointing at the created URL.

sumedhb1995:
> would like to see the png files pulled out of the repo. They can just as easily be uploaded via adding them to an issue, and then pointing at the created URL.

Updated to use URLs and removed the images. Also that's a neat tip for getting img URLs!

micahgodbolt (approved):

PR #249, opened 2021-08-09
Add react starter and cra to build

sumedhb1995:
Adds both examples to the PR build pipeline:
image


micahgodbolt:
yeah passing tests! 

Also should update learna.json, but I've got that in my upcoming PR

micahgodbolt (approved):

PR #250, opened 2021-08-09
first pass readme update for fluid-react-starter

micahgodbolt:


fluid-react-starter/README.md
```diff
@@ -74,17 +217,11 @@ export const connectionConfig: FrsConnectionConfig = useFrs
     };
 ```
 
-When just starting the app with `npm run start`, the `useFrs` value here is false and the second set of values will be used. Here, we see that our orderer and storage URLs that point to the service are directed towards 'http://localhost:7070'. The `user` object being passed into the `InsecureTokenProvider` will identify the current member's user ID and user name in the application.
```
sumedhb1995, line 77: Since we're splitting the local/FRS sections apart, could we reword this paragraph and move it up to "Run the app locally" so readers know whats happening when just `npm run start` is used?

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
 npm run start
 ```
 
-To see how this is working, take a look at `config.ts` where you will see the following values specified:
+## Modifying the model
+
+### Specify additional DDSes
+
+Inside of `src/config.ts`, in the `containerConfig` you can modify the `initialObjects` that are returned by the container.
```
sumedhb1995, line 65:
```suggestion
Inside of `src/config.ts`, you can define the `initialObjects` that are returned by the container in the `containerSchema`. 
```

sumedhb1995: Also, this variable itself in config.ts should also be renamed `containerSchema` to match the parameter name and interface

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+### Specify additional DDSes
+
+Inside of `src/config.ts`, in the `containerConfig` you can modify the `initialObjects` that are returned by the container. 
+
+To add another DSS to this list, make sure that the DDS is imported from `@fluid-experimental/fluid-framework`, select a key, and add the DDS to `initialObjects`.
+
+```ts
+import { SharedMap, SharedCounter } from '@fluid-experimental/fluid-framework';
+
+export const containerConfig = {
```
sumedhb1995, line 72:
```suggestion
export const containerSchema = {
```
The config is the first param and the schema is the second in create/getContainer calls

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+```ts
+export const setDefaultData = (fluidContainer: FluidContainer) => {
+  const { myMap, myCounter } = fluidContainer.intitialObjects;
+  myCounter.increment(1); // start at 1
+}
+```
+
+### Update the `model` to access and modify your Fluid data
+
+All of our application's business logic will be stored in `model.ts`, which is a class with access to both the `FluidContainer` and the `FrsContainerServices`.
```
sumedhb1995, line 93:
```suggestion
All of your application's business logic will be stored in `model.ts`, which is a class with access to both the `FluidContainer` and the `FrsContainerServices`. 
```

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+  const { myMap, myCounter } = fluidContainer.intitialObjects;
+  myCounter.increment(1); // start at 1
+}
+```
+
+### Update the `model` to access and modify your Fluid data
+
+All of our application's business logic will be stored in `model.ts`, which is a class with access to both the `FluidContainer` and the `FrsContainerServices`. 
+
+In the `FluidModel` class we can expose new properties and methods that can leverage any of the default DDSes or `FrsContainerServices` properties, like `audience`. These properties and methods will be used to provide a `store` of data and actions for our view to access, so keep the model focused on lower level access of the data itself.
```
sumedhb1995, line 95:
```suggestion
In the `FluidModel` class you can expose new properties and methods that can leverage any of the DDSes provided by your `FluidContainer` or `FrsContainerServices` properties, like `audience`. These properties and methods will be used to provide a `store` of data and actions for your view to access, so keep the model focused on lower level access of the data itself.
```

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+    return this.counter.value;
+  }
+  public jumpCounterFive = (): number => {
+    return this.counter.increment(5)
+  }
+```
+
+### Write custom events
+
+This demo is written to funnel all model events through the `modelChanged` event. Each event emit accepts a payload that can used to differentiate one even from another.
```
sumedhb1995, line 111:
```suggestion
This template is written to funnel all model events through the `modelChanged` event. Each event emit accepts a payload that can used to differentiate one event from another. 
```

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+  }
+  public jumpCounterFive = (): number => {
+    return this.counter.increment(5)
+  }
+```
+
+### Write custom events
+
+This demo is written to funnel all model events through the `modelChanged` event. Each event emit accepts a payload that can used to differentiate one even from another. 
+
```
sumedhb1995, line 112: May want to include a bit of context, i.e. this is an example for using a SharedCounter and its `incremented` event -> link to SharedCounter docs

fluid-react-starter/README.md
```diff
@@ -53,7 +58,145 @@ npm i
...
+    <button onClick={() => dispatch(increment())} > jump </button> </br>
+    <button onClick={() => dispatch(increment(double: true))} > BIG jump! </button> </br>
+    </div>
+  )
+}
+
+
+```
+
+
```
sumedhb1995, line 189: nit: newlines, but perhaps you're leaving them here as spacing for future content to fill out?

sumedhb1995 (approved):

fluid-react-starter/README.md
micahgodbolt: My goal here was to punt on the complexity of frs vs local configs, as the user doesn't need to even read those to start the project. Once we get down to FRS connection we talk about swapping configs due to the boolean.

PR #251, opened 2021-08-11
Update Brainstorm README to describe how the app uses the data model

timtwang:
This is a first pass update to the Brainstorm README, which aims to do the following:

- Describe how the app is initiated using Tinylicious/FrsClient
- Describe how the app uses `SharedMap` and the prefix structure to create new notes, keep track of edits made to separate notes.
- Describe how it uses audience to render user information


timtwang:
Since I will also be adding more features to demo audience in action, I will update the audience section in the next PR.

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+To add more versatility, we also incorporated the `useFrs` flag. Depending on the npm command you run (`npm run start` or `npm run start:frs`), the flag will toggle between local and remote mode using the same config format. We make use of `FrsAzFunctionTokenProvider` for running against live FRS instance since it is more secured, without exposing the tenant secret key in the client-side code whereas while running the service locally for development purpose, we make use of `InsecureTokenProvider`.
+
+```ts
+export const connectionConfig: FrsConnectionConfig = useFrs ? {
+    tenantId: "YOUR-TENANT-ID-HERE",
+    tokenProvider: new FrsAzFunctionTokenProvider("AZURE-FUNCTION-URL"+"/api/GetFrsToken", { userId: "test-user", userName: "Test User" }),
```
timtwang, line 53: The current `connectionConfig` is still using `InsecureTokenProvider` for remote mode. Perhaps it's better to change it to `FrsAzFunctionTokenProvider`?

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
 2. Create another chrome tab with `http://localhost:3000/**#1621961220840**`
 
 Now you can create notes, write text, change colors and more!
+
+## Connecting to the Service
+By configuring the `FrsConnectionConfig` that we pass into the `FrsClient` instance, we can connect to both live FRS instance by passing in the tenant ID, orderer and storage as well as using the tenant ID as "local" for running against Tinylicious for development purpose. The `FrsConnectionConfig` is defined by the `connectionConfig` constant in [Config.ts](./src/Config.ts).
+
+- Running `FrsClient` against local Tinylicious instance
```
sumedhb1995, line 35: We should have a paragraph here to guide the user to how the config gets picked up to instantiate the FrsClient in index.ts and how its used for the create/getContainer calls. Essentially this README should act as a guide that readers can use to navigate through the code, in case they want to modify it later on.

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+---
+
+- Running `FrsClient` against live FRS instance
+    - To run against live FRS Instance, tenant ID, orderer and storage URLs are required. We make use of `FrsAzFunctionTokenProvider` which takes in the Azure function URL and an IUser type object identifying the current user, thereby making an axios `GET` request call to the Azure Function. This axios call takes in the tenant ID, documentId and userID/userName as optional parameters. The Azure Function is responsible for mapping the `tenantId` to tenant key secret to generate and sign the token such that the service will accept it.
+
+To add more versatility, we also incorporated the `useFrs` flag. Depending on the npm command you run (`npm run start` or `npm run start:frs`), the flag will toggle between local and remote mode using the same config format. We make use of `FrsAzFunctionTokenProvider` for running against live FRS instance since it is more secured, without exposing the tenant secret key in the client-side code whereas while running the service locally for development purpose, we make use of `InsecureTokenProvider`.
```
sumedhb1995, line 48: I think the `npm run start:frs` command needs to get specified in the above paragraph as the way to use the FRS service. Also consider adding a bit of context for what the command does (i.e. set the environment variable that gets picked up to set the `useFrs` flag)

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+} : {
+        tenantId: "local",
+        tokenProvider: new InsecureTokenProvider("fooBar", user),
+        orderer: "http://localhost:7070",
+        storage: "http://localhost:7070",
+    };
+```
+
+## Using `SharedMap` and Prefix Structure to Update Note States
+To keep track of changes made to individual notes, the LetsBrainstorm app make use of the `SharedMap` data structure from `FluidContainer` and the prefix structure.
```
sumedhb1995, line 65: Let's add a description here of how this `SharedMap` gets specified in the containerSchema and how it matches up to the `initialObjects` returned here

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
 2. Create another chrome tab with `http://localhost:3000/**#1621961220840**`
 
 Now you can create notes, write text, change colors and more!
+
+## Connecting to the Service
+By configuring the `FrsConnectionConfig` that we pass into the `FrsClient` instance, we can connect to both live FRS instance by passing in the tenant ID, orderer and storage as well as using the tenant ID as "local" for running against Tinylicious for development purpose. The `FrsConnectionConfig` is defined by the `connectionConfig` constant in [Config.ts](./src/Config.ts).
```
micahgodbolt, line 33: Take another pass at this sentence. Bit too long and hard to follow the 2nd half.

micahgodbolt: we probably don't need to dive too deep into FRS vs tinylicious. We use this pattern in all of our examples.

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+        tokenProvider: new InsecureTokenProvider("fooBar", user),
+        orderer: "http://localhost:7070",
+        storage: "http://localhost:7070",
+    };
+```
+
+## Using `SharedMap` and Prefix Structure to Update Note States
+To keep track of changes made to individual notes, the LetsBrainstorm app make use of the `SharedMap` data structure from `FluidContainer` and the prefix structure. 
+
+The [BrainstormModel](./src/BrainstormModel.ts) defines various functions that are available to a note, including creating and deleting a note, getting likes, moving the note in the note space, changing the note text, and etc. These functions achieve their tasks by making changes to the properties associated with the note. All note properties, such as `noteId`s, `author`, `color`, `postition`, etc, are stored in a `SharedMap` as key-value pairs for easy retrieval. Now, to differentiate between different properties and notes, we make use of a prefix structure, where each key contains a prefix that indicates which property this key-value pair holds and for which note.
```
sumedhb1995, line 67: Before jumping into the description of prefix usage, lets also describe how changes applied to the map sync back with the view. It should describe the usage of the `setChangeListener` function and how it gets called in the `useEffect` hook in `NoteSpace` to update the React state holding the local notes. This loop of applying a change to the DDS, listening to the `valueChanged` event, and then using it to update local state is an important piece of the understanding for devs to take away from the example.
Also consider adding a warning to not try to modify the local state directly as it won't cause any changes for remote clients

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+To keep track of changes made to individual notes, the LetsBrainstorm app make use of the `SharedMap` data structure from `FluidContainer` and the prefix structure. 
+
+The [BrainstormModel](./src/BrainstormModel.ts) defines various functions that are available to a note, including creating and deleting a note, getting likes, moving the note in the note space, changing the note text, and etc. These functions achieve their tasks by making changes to the properties associated with the note. All note properties, such as `noteId`s, `author`, `color`, `postition`, etc, are stored in a `SharedMap` as key-value pairs for easy retrieval. Now, to differentiate between different properties and notes, we make use of a prefix structure, where each key contains a prefix that indicates which property this key-value pair holds and for which note. 
+
+```ts
+sharedMap.set(c_AuthorPrefix + noteId, newCardData.author);
+```
+While all the property prefixes are static, by attaching unique `noteId` to the end of the property prefix, we ensured that properties for each note are stored individually. Furthermore, with the use of `SharedMap`, state of each note can be updated promptly in real-time.
```
sumedhb1995, line 72: Consider walking through an example here of how one of the functions, perhaps `setNoteColor` get passed down from the model to its view component through props, and then in turn, how that action causes the `sharedMap` to trigger the `valueChanged` event. This in turn results in a new `notes` state being generated in the `syncLocalAndFluidState` function by:
1) getting the `noteIds` from the map
2) using the ids as prefixes in `createNote` to load the data for each individual note; a walkthrough of how `createNote` does this (especially for properties like `didILikeThisCalculated` that filter by user & noteId, would be great
3) applying a state update with our list of new notes. This ultimately results in the view updating.

Taking the dev through the whole loop as it is in the code may be helpful for understanding how the changes are getting applied.

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+## Using Audience to Render User Information
+The LetsBrainstorm app make use of the audience property from `FrsContainerServices` to keep track of and render all user related information. 
+
+In the [BrainstormView](./src/BrainstormView.tsx), the audience property is used to display all active users currently in the session. It is also used to retrieve current user information so the user can be assigned as the author accordingly, such as when the user creates a note.
```
sumedhb1995, line 77: Similar to above, lets include a walkthrough of how these APIs are specifically being used in this application. I.e.
what listeners are being set on the audience object and why are we calling a state update on members whenever they are triggered. Consider including a code sample here from `BrainstormView` and walking through the different audience API calls being used and adding in the background context for why they are structured that way

sumedhb1995 (commented):
Thanks for writing this up! I would recommend adding in some more details here to navigate a developer through the application. Imagine that you are seeing this code for the first time and the README is a map to figure out where to look for various functionality. It might also be helpful to add a file list at the beginning, with some descriptions for the major files (index, Config, BrainstormModel, BrainstormView, NoteSpace).
When it comes to describing the Fluid functionality specificially, I think it would be great to really step-by-step walk the reader through how calling one of the model functions results in a change on the DDS that ultimately loops back to a view state update.

brainstorm/README.md
sumedhb1995: Sounds good to me!

brainstorm/README.md
```diff
@@ -28,3 +28,50 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+---
+**NOTE**
+
+To launch the local Tinylicious service instance, run `npx tinylicious` from your terminal window
+
+---
+
+- Running `FrsClient` against live FRS instance
+    - To run against live FRS Instance, tenant ID, orderer and storage URLs are required. We make use of `FrsAzFunctionTokenProvider` which takes in the Azure function URL and an IUser type object identifying the current user, thereby making an axios `GET` request call to the Azure Function. This axios call takes in the tenant ID, documentId and userID/userName as optional parameters. The Azure Function is responsible for mapping the `tenantId` to tenant key secret to generate and sign the token such that the service will accept it.
```
sdeshpande3, line 46: Its not the `IUser` interface but rather the `FrsAzFuncUser` interface being exported from the client package

brainstorm/README.md
```diff
@@ -13,8 +13,8 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
 To run this follow the steps below:
 
 1. Run `npm install` from the brainstorm folder root
-2. Run `npm run start` to start the client
-3. Run `npx tinylicious` to start the "Tinylicious" test service
+2. Run `npx tinylicious` to start the "Tinylicious" test service
```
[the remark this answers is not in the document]
timtwang: good call. I added more context explaining each server

brainstorm/README.md
timtwang: I agree, but I also feel like having more context/explanation won't hurt as long as it doesn't divert from the topic too much? It also easier for developer to just get a brief overview and if they want to read more about FRS vs Tinylicious they can go and read the docs for it?

timtwang: Another solution is I can remove the 2 bullet points explaining FRS vs Tinylicious and just include the code snippet of what the `FrsConnectionConfig` should look like.

brainstorm/README.md
timtwang: I'm not seeing a `FrsAzFuncUser` type, instead, I see a ` Pick<FrsMember, "userId" | "userName">`, which I think extracts the `userId` and `userName` from the `FrsMember` type.

brainstorm/README.md
timtwang: good point, I separated the `getting started` section into 2, local and remote modes.

brainstorm/README.md
timtwang: added a description of `SharedMap` and how it relates to `containerSchema` and `initialObjects` 😃

brainstorm/README.md
sumedhb1995: Yep, instead of `IUser` listing out the `"userId" | "userName" | "additionalDetails"` parameters as the ones they need to pass in would be great

brainstorm/README.md
timtwang: Good point. I added more explanation for the loop. Let me know what you think 😀

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
-To run this follow the steps below:
+### Follow the steps below to run this in local mode (Tinylicious):
 
 1. Run `npm install` from the brainstorm folder root
-2. Run `npm run start` to start the client
-3. Run `npx tinylicious` to start the "Tinylicious" test service
-4. Navigate to `http://localhost:3000` in a browser tab
+2. Run `npx tinylicious` to start the "Tinylicious" test service
+3. Run `npm run start` to start the client
+5. Navigate to `http://localhost:3000` in a browser tab
```
sumedhb1995, line 18:
```suggestion
4. Navigate to `http://localhost:3000` in a browser tab
```

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
+|:---------------------------------------------------------------------------------------------------------|
+| Tinylicious is a local, self-contained test service. By running `npx tinylicious` from your terminal window will launch the Tinylicious server. The server will need to be started first in order to provide the ordering and storage requirement of Fluid runtime.                                                         |
+
+<br />
+
+### Follow the steps below to run this in remote mode (Routerlicious):
+
+1. Run `npm install` from the brainstorm folder root
+2. Run `npm run start:frs` to start the "Routerlicious" test service
+5. Navigate to `http://localhost:3000` in a browser tab
```
sumedhb1995, line 32:
```suggestion
3. Navigate to `http://localhost:3000` in a browser tab
```

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
-4. Navigate to `http://localhost:3000` in a browser tab
+2. Run `npx tinylicious` to start the "Tinylicious" test service
+3. Run `npm run start` to start the client
+5. Navigate to `http://localhost:3000` in a browser tab
+
+<br />
+
+| :memo: NOTE                                                                                              |
+|:---------------------------------------------------------------------------------------------------------|
+| Tinylicious is a local, self-contained test service. By running `npx tinylicious` from your terminal window will launch the Tinylicious server. The server will need to be started first in order to provide the ordering and storage requirement of Fluid runtime.                                                         |
```
sumedhb1995, line 24:
```suggestion
| Tinylicious is a local, self-contained test service. Running `npx tinylicious` from your terminal window will launch the Tinylicious server. The server will need to be started first in order to provide the ordering and storage requirement of Fluid runtime.                                                         |
```

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
+
+<br />
+
+### Follow the steps below to run this in remote mode (Routerlicious):
```
sumedhb1995, line 28: Let's refer to Routerlicious here and below as Azure Fluid Relay service, since that's the "official" naming and that's what devs will see when onboarding from the Azure portal. Otherwise, users may get confused as to what Routerlicious refers to 😄

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
+
+1. Run `npm install` from the brainstorm folder root
+2. Run `npm run start:frs` to start the "Routerlicious" test service
```
sumedhb1995, line 31:
```suggestion
2. Run `npm run start:frs` to connect to the Azure Fluid Relay service
```

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
...
+5. Navigate to `http://localhost:3000` in a browser tab
+
+<br />
+
+| :memo: NOTE                                                                                              |
+|:---------------------------------------------------------------------------------------------------------|
+| Routerlicious is a main composed server definition that pulls together multiple micro-services that provide the ordering and storage requirement of Fluid runtime. By running `npm run start:frs` from your terminal window, the environment variable `REACT_APP_FLUID_CLIENT` will be set first, which will be picked up by the `useFrs` flag, and `FrsConnectionConfig` will use the remote mode config format. Then, the command will starts the server.                                                                            |
```
sumedhb1995, line 38:
```suggestion
| Azure Fluid Relay service is a deployed service implementation that pulls together multiple micro-services that provide the ordering and storage requirement of Fluid runtime. By running `npm run start:frs` from your terminal window, the environment variable `REACT_APP_FLUID_CLIENT` will be set first, which will be picked up by the `useFrs` flag, and `FrsConnectionConfig` will use the remote mode config format. Please use the values provided as part of the service onboarding process to fill in this configuration. Then, the command will connect to your service instance.                                                                            |
```

sumedhb1995: The command just connects the developer to their already deployed instance. The service itself gets started when the developer creates it from the Azure portal (like it was demoed at readout) and then they get the id, key, and orderer and storage URLs from there

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+- Running `FrsClient` against local Tinylicious instance
+    - To run against our local Tinylicious instance, we pass the `tenantId` as "local" and make use of `InsecureTokenProvider`. The `InsecureTokenProvider` requires we pass in two values to its constructor, a key string, which can be anything since we are running it locally, and an IUser type object identifying the current user. For running the instance locally, the orderer and storage URLs would point to the Tinylicious instance on the default values of `http://localhost:7070`.
+
+- Running `FrsClient` against live FRS instance
+    - To run against live FRS Instance, tenant ID, orderer and storage URLs are required. We make use of `FrsAzFunctionTokenProvider` which takes in the Azure function URL and an optional `"userId" | "userName" | "additionalDetails"` type object identifying the current user, thereby making an axios `GET` request call to the Azure Function. This axios call takes in the tenant ID, documentId and userID/userName as optional parameters. The Azure Function is responsible for mapping the `tenantId` to tenant key secret to generate and sign the token such that the service will accept it.
```
sumedhb1995, line 61: Just a heads up that the `FrsClient` and `FrsAzFunctionTokenProvider` references will need to get updated to `AzureClient` and `AzureFunctionTokenProvider` once we have our next release due to the renaming. Its fine for now though as it matches the code

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+} : {
+        tenantId: "local",
+        tokenProvider: new InsecureTokenProvider("fooBar", user),
+        orderer: "http://localhost:7070",
+        storage: "http://localhost:7070",
+    };
+```
+
+## Using `SharedMap` and Prefix Structure to Update Note States
+To keep track of changes made to individual notes, the LetsBrainstorm app make use of the `SharedMap` data structure from `FluidContainer` and the prefix structure.
```
sumedhb1995, line 80:
```suggestion
To keep track of changes made to individual notes, the LetsBrainstorm app makes use of the `SharedMap` data structure from `FluidContainer` and a prefix-based structuring of our data for tracking different properties.
```

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+3. Apply a state update with our list of new notes.
+    - With our newly generated and updated list of new notes, we call `setNotes` to update the React state. This updated React state will propagate the changes to all remote clients, resulting in the view updating.
+
+## Using Audience to Render User Information
+The LetsBrainstorm app make use of the `audience` property from `FrsContainerServices` to keep track of and render all user related information. 
+
+In the [BrainstormView](./src/BrainstormView.tsx), the audience property is used to achieve 2 tasks, display all active users currently in the session, and retrieve current user information so the user can be assigned as the author accordingly, such as when the user creates a note.
+
+Similar to how `BrainstormModel` works in [NoteSpace.tsx](./src/view/NoteSpace.tsx), the member values of the audience property is also being tracked as a React state.
```
sumedhb1995, line 117:
```suggestion
Similar to how `BrainstormModel` works in [NoteSpace.tsx](./src/view/NoteSpace.tsx), the member values of the audience property are also being tracked in a React state.
```

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+    storage: "ENTER-STORAGE-URL-HERE",
+} : {
+        tenantId: "local",
+        tokenProvider: new InsecureTokenProvider("fooBar", user),
+        orderer: "http://localhost:7070",
+        storage: "http://localhost:7070",
+    };
+```
+
+## Using `SharedMap` and Prefix Structure to Update Note States
```
sumedhb1995, line 79: I **love** the writing in this section!! Awesome walkthrough!! 😍

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+```ts
+const setMembersCallback = React.useCallback(() => setMembers(
+    Array.from(
+      audience.getMembers().values()
+    )
+  ), [setMembers, audience]);
+```
+
+The `setMembersCallback()` retreives a list of all the active members and convert it to an array. This array is then used by `setMembers` to update the React state of active users when new clients join or leave the session.
```
sumedhb1995, line 133:
```suggestion
The `setMembersCallback()` retrieves a list of all the active members and convert it to an array. This array is then used by `setMembers` to update the React state of active users when new clients join or leave the session.
```

brainstorm/README.md
```diff
@@ -28,3 +50,92 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+Now, audience also has a `getMyself()` property to get the current client as a member. By passing this into the view, [NoteSpace.tsx](./src/view/NoteSpace.tsx), as props, this allows the user to be assigned as author whenever the user creates a note.
```
sumedhb1995, line 135:
```suggestion
Now, audience also has a `getMyself()` property to get the current client as a member. Passing this into the view, [NoteSpace.tsx](./src/view/NoteSpace.tsx), as props, allows the user to be assigned as author whenever the user creates a note.
```

sumedhb1995 (approved):
Awesome writing @timtwang!!! This is a fantastic walkthrough of the app. Just left a few small comments regarding terminology/grammar but content-wise, this is solid. Nice work! :shipit:

brainstorm/README.md
sumedhb1995: Yeah, I think providing some context for the devs in the example READMEs would be helpful since we don't know what other documentation/examples they are coming from. If this is say the first example they're seeing, having the distinction between local vs remote, and how to set up the config for each would be helpful for them to understand what's happening in the code. Also, I like the rephrased content 😄

brainstorm/README.md
```diff
@@ -28,3 +50,124 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+
+- Syncing `SharedMap` and View data
+    - The [BrainstormModel](./src/BrainstormModel.ts) defines various functions that are available to a note, including creating and deleting a note, getting likes, moving the note in the note space, changing the note text, and etc. These functions achieve their tasks by making changes to the properties associated with the note. All note properties, such as `noteId`s, `author`, `color`, `postition`, etc, are stored in a `SharedMap` as key-value pairs for easy retrieval. Now, syncing our Fluid and View data requires that we set up an event listener, which mean we need a `useEffect()` hook, defined in [NoteSpace.tsx](./src/view/NoteSpace.tsx).
+
+    To sync the data, we created a `syncLocalAndFluidState()` function, called that function once to initialize the data, and then keep listening for the `SharedMap` "valueChanged" event in `setChangeListener()`, and fire the function again each time. Now React will handle updating the view each time the new `notes` state is modified.
```
sumedhb1995, line 87: The rewording here from the earlier commit looks great and I like the inclusion of the code sample. Could we move it up to be above the "To sync the data..." paragraph so that readers can get the code and then follow it up with the walkthrough of it?

brainstorm/README.md
timtwang: thanks! I got a lot of inspirations from other demo READMEs 😛

brainstorm/README.md
timtwang: sounds good!!

brainstorm/README.md
timtwang: Yup I moved it up. The rewording was inspired by @micahgodbolt awesome CRA README! 😄

brainstorm/README.md
timtwang: yeah, I was a bit confused on when to use which term myself too 😅

brainstorm/README.md
```diff
@@ -10,12 +10,34 @@ This application was shown during a [Microsoft Build session](https://aka.ms/OD5
 
 ## Getting Started
 
-To run this follow the steps below:
+### Follow the steps below to run this in local mode (Tinylicious):
```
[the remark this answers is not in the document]
timtwang: Removed the header symbol 🙂

brainstorm/README.md
timtwang: I think I have yet to fully understand how Routerlicious works, but this clarifies a lot! 😃

timtwang:
> Awesome writing @timtwang!!! This is a fantastic walkthrough of the app. Just left a few small comments regarding terminology/grammar but content-wise, this is solid. Nice work! :shipit: 

Really appreciated all the detailed comments. It was super helpful!! 😀

brainstorm/README.md
```diff
@@ -52,13 +52,45 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+    const client = new FrsClient(connectionConfig);
+
+    const frsResources = isNew
+        ? await client.createContainer({ id: containerId }, containerSchema)
+        : await client.getContainer({ id: containerId }, containerSchema);
+    ...
+}
+```
+
+Since `start()` is an async function, we'll need to await for the initialObjects to be returned. Once returned, each initialObjects key will point to a connected data structure as defined in the schema.
```
sumedhb1995, line 87:
```suggestion
Since `start()` is an async function, we'll need to await for the `initialObjects` to be returned. Once returned, each `initialObjects` key will point to a connected data structure as defined in the schema.
```

sumedhb1995: nit

brainstorm/README.md
```diff
@@ -52,13 +52,45 @@ You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' t
...
+Now, before you can access any Fluid data, you need to define your container schema after creating a configured `FrsClient` using `FrsConnectionConfig`.
 
-- Running `FrsClient` against live FRS instance
-    - To run against live FRS Instance, tenant ID, orderer and storage URLs are required. We make use of `FrsAzFunctionTokenProvider` which takes in the Azure function URL and an optional `"userId" | "userName" | "additionalDetails"` type object identifying the current user, thereby making an axios `GET` request call to the Azure Function. This axios call takes in the tenant ID, documentId and userID/userName as optional parameters. The Azure Function is responsible for mapping the `tenantId` to tenant key secret to generate and sign the token such that the service will accept it.
+- `containerSchema`, also defined in [Config.ts](./src/Config.ts), is going to include a string `name` and a collection of the data types our application will use.
```
sumedhb1995, line 59: May want to also include a code sample from Config.ts of the `containerSchema` here just to have them side-by-side

timtwang: added!

PR #272, opened 2021-08-24
Additional functionalities to further demo Audience in action

timtwang:
This purpose of this PR aims to update Brainstorm with sensible audience-usage examples. To do this, the following features are added:

1. `userId` is now replaced with `userName` in the UI so as to be more human readable.
userName

2. Tooltip of the note persona will display "Created by {author name}". If the author is also the user, the tooltip will instead display "Created by you".
createdBy

3. The user last edited the note will be shown in the footer of the note. Similar to the tooltip, it will display "Last edited by {author name}" or "Last edited by you" depending on if the user is the last one to edit the note. The feature also waits 2 seconds before actually updating the last edited user so as to prevent flickering.
last_edited_by


brainstorm/src/view/NoteFooter.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <div style={{ flex: 1 }}>
+      <TextField
+        styles={{ fieldGroup: { background: ColorOptions[color].light}, field: { color: "grey"}}}
+        borderless
+        readOnly={true}
+        resizable={false}
+        autoAdjustHeight
+        value={"Last edited by " + lastEditedMemberName}
```
ChumpChief, line 29: Nit: here and elsewhere prefer template strings over concatenation.

brainstorm/src/view/NoteSpace.tsx
```diff
@@ -30,9 +30,18 @@ export function NoteSpace(props: NoteSpaceProps) {
       setNotes(noteDataArr);
     };
 
+    // Timer to refresh so as to update the "last edited" feature when there is no edit
+    const timer = setInterval(() => {
```
ChumpChief, line 34: This is not a great strategy -- we should have sufficient signal for when the last-edited changes to set up a one-time timer to update the UI rather than a continuously spinning interval (see above note).

brainstorm/README.md
```diff
@@ -199,10 +197,299 @@ React.useEffect(() => {
...
+
+1. displaying all current active users
+2. displaying author name in persona tooltip 
+3. displaying like and the note's liked users
+4. displaying the note's last edited user
+
+### 1. Displaying All Current Actuve Users
+To display all active members in the session, we are using the `Facepile` element from `@fluentui/react`. This element will display all the active members in their respective personas. The persona will display the image if the user has one, otherwise, it will display the first letter of their names. Now, in order to use the `Facepile` element, we first pass the `member` variable into our `Header` element, defined [here](./src/view/Header.tsx), as argument.
+
+```ts
```
ChumpChief, line 216: I read through the full README but it's really verbose, not sure how useful this is as a learning tool.  I also note that although this gives a blow-by-blow of the code, there are few to none code comments reflecting similar info.  Perhaps consider significantly reducing the README in favor of in-code documentation of the bits that matter most.

ChumpChief (commented):
Lots of nits and proofreading, but main feedback is to try to eliminate the interval.

brainstorm/src/view/NoteFooter.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ColorOptions, DefaultColor } from "./Color";
+
+export type NoteFooterProps = { currentUser : FrsMember } & Pick<NoteData,  "lastEdited" | "color">;
+
+export function NoteFooter(props: NoteFooterProps) {
+  const { currentUser, lastEdited, color = DefaultColor } = props;
+  let lastEditedMemberName;
+
+  // To prevent flickering, wait for 2s to ensure no one else is editing the note
+  if((Date.now() - lastEdited.time) >= 2000) {
```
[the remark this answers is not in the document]
timtwang: @ChumpChief  Didn't really have much experience with caching in React, but I finally got it hashed out after a day and a half!! 🎊😫. Although not caching the timer, but each note has their own timer, when the dirty flag is set, it will clear the old timer and kick start a new one; hence, only 1 timer for each note at a time. @sumedhb1995 and I did try to find a more native/intuitive way of doing such update by including a `isDirtyMap` in the `BrainstormModel`, but that would require a major refactoring in multiple view components. Anyway, after a few trial and errors, we figured the current implementation is good enough for the scope and purpose of this app. However, we also concluded that we need to incorporate a better way of updating local states in the future (@sumedhb1995 can explain it further).

Despite how long it took, all of this was a major learning experience though and definitely worth the deep dive!! 😃

brainstorm/src/BrainstormModel.ts
```diff
@@ -13,7 +14,7 @@ export type BrainstormModel = Readonly<{
   CreateNote(noteId: string, myAuthor: FrsMember): NoteData;
   MoveNote(noteId: string, newPos: Position): void;
   SetNote(noteId: string, newCardData: NoteData): void;
-  SetNoteText(noteId: string, noteText: string): void;
+  SetNoteText(noteId: string, noteText: string, lastEdited: FrsMember): void;
```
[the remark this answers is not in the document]
ChumpChief: Yeah that's good feedback - we shouldn't be storing any more than we strictly want to track.  It's benign here but paring this down to just the `userName` seems like a good idea.

brainstorm/src/view/NoteFooter.tsx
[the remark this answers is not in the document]
ChumpChief: I hate to say it but I think I prefer the interval option here -- between `setInterval` and `refreshView` I'd prefer `setInterval`.  I see what you mean in terms of the refactoring being a bigger project (though I think worthwhile in the future).  If/when that refactor takes place, a timeout-based solution like this would be good (and probably a lot easier, since you'd likely not need to use refs).

brainstorm/README.md
```diff
@@ -199,10 +197,299 @@ React.useEffect(() => {
 
 To sync the data, we created a `setMembersCallback()` function, which retrieves a list of all the active members and convert it to an array, then have a listener keep listening for the "membersChanged" event, and fire the function each time. Now React will handle updating the view each time the new `members` state is modified.
 
-Now, audience also has a `getMyself()` property to get the current client as a member. Passing this into the view, [NoteSpace.tsx](./src/view/NoteSpace.tsx), as props, this allows the user to be assigned as author whenever the user creates a note.
+Now, audience also has a `getMyself()` property to get the current client as a member. Passing this into the view as props will allows the user information to be displayed or processed when the user performs different note operations (creating a note, liking a note, and editing a note).
```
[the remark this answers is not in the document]
timtwang: I think we all agree that this README is a bit verbose, especially in the audience section. After discussing with @heliocliu, I'm going to take out the explanation for all audience features and instead just focus on the "last edited by" as an walk-through example, similar to `sharedmap` in the previous section.

brainstorm/src/BrainstormModel.ts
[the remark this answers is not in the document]
timtwang: I agree with saving only the specific info we need given the purpose of this app is to educate developers. I also feel like we should mention object doesn't get stored in the DDS map in the same way as it would be stored in a regular map. Perhaps in the `SharedMap` section as a note/warning box?

brainstorm/README.md
timtwang: I also feel like it's very verbose, but only in the audience section. I decided to take out explaining all audience features and instead just dive into the "last edited by" feature as a walk-through example

brainstorm/src/BrainstormModel.ts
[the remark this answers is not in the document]
heliocliu: Sounds like a good plan as a followup

brainstorm/src/view/NoteFooter.tsx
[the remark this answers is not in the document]
sumedhb1995: The main takeaway here ended up being that it was difficult to have both local state and remote state that was governed by Fluid. Anytime that we had the React component logic working correctly to update an `isDirty` state for each note, it would end up getting overwritten since the note state is being completely recreated on each `valueChanged` event on the map. This meant that any local state changes would not get preserved and end up resorting to local values 😢 I believe to change this, we need to a key-based state update instead of regenerating the entire state each time

brainstorm/README.md
```diff
@@ -170,9 +170,9 @@ To summarize how these 2 components work together seamlessly, let's take `setNot
     - With our newly generated and updated list of new notes, we call `setNotes` to update the React state. This updated React state will propagate the changes to all remote clients, resulting in the view updating.
 
 ## Using Audience to Render User Information
-The LetsBrainstorm app make use of the `audience` property from `FrsContainerServices` to keep track of and render all user related information. Namely, we are using `audience` to create a list of active members in the session and retrieve user information.
+The LetsBrainstorm app make use of the `audience` property from `FrsContainerServices` to keep track of and render all user related information.
```
tylerbutler, line 173:
```suggestion
The LetsBrainstorm app makes use of the `audience` property from `FrsContainerServices` to keep track of and render all user related information.
```

brainstorm/src/view/NoteFooter.tsx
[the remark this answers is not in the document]
ChumpChief: Yeah I took a further look and saw the complexity you're talking about.  I messaged Tim earlier - I'd say go with whichever for initial checkin and file an issue to revisit later, I think it's better to have something in that we can iterate on.

brainstorm/src/BrainstormModel.ts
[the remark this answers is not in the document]
timtwang: Follow-up now in https://github.com/microsoft/FluidFramework/issues/7303

brainstorm/src/view/NoteFooter.tsx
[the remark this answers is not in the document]
timtwang: > Overall I'd like a lot of the words to be moved to comments in the code, with the readme focused on the big picture. That might be more than you can tackle at the moment.

I created an issue (https://github.com/microsoft/FluidFramework/issues/7303) to add more code . 😃

timtwang:
> Overall I'd like a lot of the words to be moved to comments in the code, with the readme focused on the big picture. That might be more than you can tackle at the moment.

I created an issue to add more code comment https://github.com/microsoft/FluidFramework/issues/7303 😀

brainstorm/src/view/NoteFooter.tsx
```diff
@@ -0,0 +1,33 @@
...
+    lastEditedMemberName = currentUser?.userName === lastEdited.member.userName? "you" : lastEdited.member.userName;
+  }
+  else {
+    lastEditedMemberName = "...";
+  }
+
+  return (
+    <div style={{ flex: 1 }}>
```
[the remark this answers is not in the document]
timtwang: tracking the issue here. https://github.com/microsoft/FluidFramework/issues/7303

brainstorm/src/view/NoteFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if(!isDirty.current) {
+    lastEditedMemberName = currentUser?.userName === lastEdited.userName ? "you" : lastEdited.userName;
+  }
+  else {
+    lastEditedMemberName = "...";
+  }
+
+  return (
+    <div style={{ flex: 1 }}>
+      <TextField
```
micahgodbolt, line 50: Is there a strong need here for a Textfield? If you're just displaying text here than this is not the most accessible choice.

timtwang: I originally chose TextField because it allows for onChange but it turns out I didn't need it. Another reason was so that I can match the `NoteBody` attributes. I believe the only alternative FluentUI offers here is the Label, but I don't think it's suitable in this scenario.

brainstorm/src/view/NoteFooter.tsx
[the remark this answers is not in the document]
timtwang: I reverted back to the interval approach and created a work item for refactoring that can be tracked here https://github.com/microsoft/FluidFramework/issues/7303

brainstorm/src/view/NoteFooter.tsx
micahgodbolt: [Text](https://developer.microsoft.com/en-us/fluentui#/controls/web/text) should be fine.

timtwang: Replaced with Text! 😀

heliocliu:
> > Overall I'd like a lot of the words to be moved to comments in the code, with the readme focused on the big picture. That might be more than you can tackle at the moment.
> 
> I created an issue to add more code comment [microsoft/FluidFramework#7303](https://github.com/microsoft/FluidFramework/issues/7303) 😀

@tylerbutler Asking since I have no clue or if it even matters, issues for examples should be in the examples repo I'd assume?

timtwang:
> > > Overall I'd like a lot of the words to be moved to comments in the code, with the readme focused on the big picture. That might be more than you can tackle at the moment.
> > 
> > 
> > I created an issue to add more code comment [microsoft/FluidFramework#7303](https://github.com/microsoft/FluidFramework/issues/7303) 😀
> 
> @tylerbutler Asking since I have no clue or if it even matters, issues for examples should be in the examples repo I'd assume?

I tried creating an issue on the examples repo but it redirected me to the main repo 🤔

brainstorm/README.md
```diff
@@ -199,10 +197,94 @@ React.useEffect(() => {
 
 To sync the data, we created a `setMembersCallback()` function, which retrieves a list of all the active members and convert it to an array, then have a listener keep listening for the "membersChanged" event, and fire the function each time. Now React will handle updating the view each time the new `members` state is modified.
 
-Now, audience also has a `getMyself()` property to get the current client as a member. Passing this into the view, [NoteSpace.tsx](./src/view/NoteSpace.tsx), as props, this allows the user to be assigned as author whenever the user creates a note.
+The audience object also has a `getMyself()` function that returns the current client as a member. This is passed in as a view prop so that the user information to be displayed or processed when the user performs different note operations (creating a note, liking a note, and editing a note).
```
heliocliu, line 200:
```suggestion
The audience object also has a `getMyself()` function that returns the current client as a member. This is passed in as a view prop so that the user information can be displayed or processed when the user performs different note operations (creating a note, liking a note, and editing a note).
```

brainstorm/README.md
```diff
@@ -199,10 +197,94 @@ React.useEffect(() => {
...
 
 ```ts
 const authorInfo = audience.getMyself();
 ```
 
+With `members` and `authorInfo` defined, we can use these to achieve several tasks:
+
+1. displaying all current active users
+2. displaying author name in persona tooltip 
+3. displaying like and the note's liked users
```
heliocliu, line 210:
```suggestion
3. displaying likes for the note
```

brainstorm/README.md
```diff
@@ -199,10 +197,94 @@ React.useEffect(() => {
...
+
+```ts
+const setText = (text: string) => {
+    model.SetNoteText(note.id, text, props.author);
+};
+```
+
+As seen in here in [NoteSpace.tsx](./src/view/NoteSpace.tsx), the `setText()` function calls the `SetNoteText()` function from [BrainstormModel](./src/BrainstormModel.ts), passing in `props.author`, which is the `authorInfo` passed to the `NoteSpace` element as argument. Similar to `onLike()` mentioned previously, the `setText()` function defined here is passed into [NoteBody.tsx](./src/view/NoteBody.tsx) as props where it will be called when the text changes as seen below.
```
heliocliu, line 222:
```suggestion
As seen here and in [NoteSpace.tsx](./src/view/NoteSpace.tsx), the `setText()` function calls the `SetNoteText()` function from [BrainstormModel](./src/BrainstormModel.ts), passing in `props.author`, which is the `authorInfo` passed to the `NoteSpace` element. The `setText()` function is passed into [NoteBody.tsx](./src/view/NoteBody.tsx) where it will be called when the text changes as seen below.
```

brainstorm/README.md
```diff
@@ -199,10 +197,94 @@ React.useEffect(() => {
...
+
+Going back to `SetNoteText()` from [BrainstormModel](./src/BrainstormModel.ts), we can see in the definition below that we are not only updating the last edited member but also giving it a timestamp of when it was last edited.
+
+```ts
+const SetNoteText = (noteId: string, noteText: string, lastEditedMember: FrsMember) => {
+    sharedMap.set(c_TextPrefix + noteId, noteText);
+    sharedMap.set(c_LastEditedPrefix + noteId, { member: lastEditedMember, time: Date.now() });
+  };
+```
+The reason for adding a timestamp is because given that Fluid updates so quickly, in the event where multiple users are editing the same note, we want to wait a little bit after all changes are done before displaying the last edited user. By doing this, we can prevent the text field from flickering with user names, which could potentially be distracting.
```
heliocliu, line 249:
```suggestion
In the event where multiple users are editing the same note, we want to wait a little bit after all changes are done before displaying the last edited user. By doing this, we can prevent the text field from flickering with user names.
```

heliocliu:
Looks like you addressed storing the FrsMember in the SharedMap here instead of as part of the follow up issue.  Those changes should be reflected in the README code snippets too, but I think the README generally needs some editing around clarity beyond the verbosity from before so it can all be handled together in the follow up.

heliocliu (approved):

PR #288, opened 2021-09-08
update fluid-react-starter to new container approach

micahgodbolt:


fluid-react-starter/README.md
```diff
@@ -64,13 +59,12 @@ npm run start
 
 Inside of `src/config.ts`, you can define the `initialObjects` that are returned by the container in the `containerSchema`. 
 
-To add another DSS to this list, make sure that the DDS is imported from `@fluid-experimental/fluid-framework`, select a key, and add the DDS to `initialObjects`.
+To add another DSS to this list, make sure that the DDS is imported from `fluid-framework`, select a key, and add the DDS to `initialObjects`.
```
tylerbutler, line 62:
```suggestion
To add another DDS to this list, make sure that the DDS is imported from `fluid-framework`, select a key, and add the DDS to `initialObjects`.
```

tylerbutler (approved):

sumedhb1995:
Are we planning on not having the fluid-react-starter be connected to Azure Fluid Relay service at all? I'm good for moving the deployment instructions out, but I feel like we should still have the usage of the AzureClient in here

micahgodbolt:
@sumedhb1995 I'm happy to bring it back sometime, but for now I'm happy reducing the number of places we use AzureClient until we have all of the azure specific docs in place...then we can go back and see how we can make converting from tinylicious to azure as easy as possible.

PR #299, opened 2021-09-15
Add Angular starter code for example with Epoch time

scottn12:
This PR adds a simple example for using Fluid with Angular CLI. This application has a single button to get the UNIX Epoch time and uses a SharedMap to store and synchronize the time between the clients.


fluid-angular-starter/README.md
```diff
@@ -0,0 +1,45 @@
+# fluid-angular-starter
+
+## About this repo
+This repo contains an Angular starter template to demonstrate how to use Fluid Framework in a simple Angular application. To use this
```
heliocliu, line 4: it looks like the last iteration removed some directions from this file.  is the plan still to cover this in a followup pr?  if so it may be better either to leave this as it was or remove the instructions here, since as is it looks like it's missing some crucial steps

micahgodbolt:
This is just opinion (as i'm not an angular dev), but when I wrote the create-react-app demo, I stripped out a large chunk of the default files that did not have to do with Fluid. I feel it makes it easier for someone to come into the repo and see what changes are required to go from setup script to working Fluid app. 

Is this something we can do with the angular approach? We don't expect someone to clone this demo, rather to follow the same steps of "run setup script, change these few files".  This also means not calling this "starter", but fluid-angular-demo.

micahgodbolt:
The "starter" package in this repo is a more full blown app with a ton of boilerplate setup to get someone building a production app more quickly.....not as a demo you can quickly parse.

fluid-angular-starter/src/polyfills.ts
```diff
@@ -0,0 +1,65 @@
...
+ *  with the following flag, it will bypass `zone.js` patch for IE/Edge
+ *
+ *  (window as any).__Zone_enable_cross_context_check = true;
+ *
+ */
+
+/***************************************************************************************************
+ * Zone JS is required by default for Angular itself.
+ */
+import 'zone.js';  // Included with Angular CLI.
```
heliocliu, line 60: is this required regardless of browser?  if so, is there a way to move this into the package.json dependencies in a way that this extra file isn't needed?

skylerjokiel:
We need to add copyright comments at the top of js/ts files.

```js
/*!
 * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
 * Licensed under the MIT License.
 */
```

heliocliu:
> This is just opinion (as i'm not an angular dev), but when I wrote the create-react-app demo, I stripped out a large chunk of the default files that did not have to do with Fluid. I feel it makes it easier for someone to come into the repo and see what changes are required to go from setup script to working Fluid app.
> 
> Is this something we can do with the angular approach? We don't expect someone to clone this demo, rather to follow the same steps of "run setup script, change these few files". This also means not calling this "starter", but fluid-angular-demo.

Agree on removing as much of the unnecessary stuff as we can.  @micahgodbolt is there also any precedent on the tests?  Is that something we want as part of this level of demo?

micahgodbolt:
As for tests, that might be a better addition to a larger, more opinionated "starter" repo. I believe the goal of this demo is to have a "minimum viable demo" using angular. 

I think it comes down to acknowledging that we should have a minimum viable "demo" first, and then consider a more robust, opinionated "starter" later

fluid-angular-starter/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@angular/compiler": "~12.2.0",
+    "@angular/core": "~12.2.0",
+    "@angular/forms": "~12.2.0",
+    "@angular/platform-browser": "~12.2.0",
+    "@angular/platform-browser-dynamic": "~12.2.0",
+    "@angular/router": "~12.2.0",
+    "@fluidframework/tinylicious-client": "^0.47.0",
+    "concurrently": "^6.2.1",
+    "fluid-framework": "^0.47.0",
+    "rxjs": "~6.6.0",
```
skylerjokiel, line 26: Is this import required?

fluid-angular-starter/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@angular/animations": "~12.2.0",
+    "@angular/common": "~12.2.0",
+    "@angular/compiler": "~12.2.0",
+    "@angular/core": "~12.2.0",
+    "@angular/forms": "~12.2.0",
+    "@angular/platform-browser": "~12.2.0",
+    "@angular/platform-browser-dynamic": "~12.2.0",
+    "@angular/router": "~12.2.0",
+    "@fluidframework/tinylicious-client": "^0.47.0",
+    "concurrently": "^6.2.1",
```
skylerjokiel, line 24: `concurrently` should be in `devDependencies`

skylerjokiel (changes requested):
@scottn12 this is awesome! I would love to capture how you got to this state. Did you start from an empty Angular template or something else. I've left some high level comments but I would love to get this in and continue to iterate on how we tell our angular story!

scottn12:
Follow up notes:

- We should add a note to the React starter code README to indicate that Node must be installed on your machine (and give instructions on how to install)

fluid-angular-starter/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,9 @@
+# Microsoft Open Source Code of Conduct
```
heliocliu, line 1: i think what skyler meant about these was that you don't need them at all.  they already exist in the parent directory (the root of this repo), and apply to the entirety of the repo

scottn12: Oh okay, I wasn't sure if this was only in reference to the Contributing and Trademarks sections in the README. I should remove the `LICENSE,` `CODE_OF_CONDUCT.md`, and `SECURITY.md` files?

fluid-angular-starter/README.md
```diff
@@ -0,0 +1,182 @@
+# fluid-angular-starter
+
+## About this repo
+
+This repo is a Fluid starter template that was created to answer the question "how do I create a Fluid app that is more complex than Hello World?" To answer this question this repo makes the following assumptions:
+
+1. You want to use Angular CLI for your view
```
heliocliu, line 7: is the CLI used for the view/should this just be "Angular"?

scottn12: I wanted to clarify somewhere that it is Angular CLI and not AngularJS. Do you think I should just leave it out since we generally just call it Angular everywhere else?

heliocliu: I think just Angular is probably better then.  The intent here is that a developer wants to run on Angular, and to that extent it's not super important what you use to create your view as long as it runs on the expected framework

fluid-angular-starter/CODE_OF_CONDUCT.md
heliocliu: yup

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  fluidSharedObjects: FluidDataModel | undefined;
+  localTimestamp: TimestampDataModel | undefined;
+
+  async ngOnInit() {
+    this.fluidSharedObjects = await this.getFluidData();
+    this.syncData();
+  }
+
+  async getFluidData() {
+
```
heliocliu, line 29: nit: remove buffer whitespace
```suggestion
```

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,77 @@
...
+      const id = await container.attach();
+      location.hash = id;
+    }
+    else {
+      ({ container } = await client.getContainer(containerId, containerSchema));
+    }
+
+    // TODO 3: Return the Fluid timestamp object.
+    return container.initialObjects as any as FluidDataModel;
+
```
heliocliu, line 50:
```suggestion
```

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.fluidSharedObjects = await this.getFluidData();
+    this.syncData();
+  }
+
+  async getFluidData() {
+
+    // TODO 1: Configure the container.
+    const client = new TinyliciousClient();
+    const containerSchema = {
+      initialObjects: { sharedTimestamp: SharedMap }
```
heliocliu, line 33: given you've defined it earlier, would it be better to use `FluidDataModel` as the type here instead

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // TODO 3: Return the Fluid timestamp object.
+    return container.initialObjects as any as FluidDataModel;
+
+  }
+
+  syncData() {
+    if (this.fluidSharedObjects) {
+
```
heliocliu, line 53:
```suggestion
```

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,75 @@
...
+      const { sharedTimestamp } = this.fluidSharedObjects;
+      const updateLocalTimestamp = () => this.localTimestamp = { time: sharedTimestamp.get("time") };
+      updateLocalTimestamp();
+
+      // TODO 5: Register handlers.
+      sharedTimestamp.on("valueChanged", updateLocalTimestamp);
+
+      // TODO 6: Delete handler registration when the React App component is dismounted.
+      return () => { sharedTimestamp.off("valueChanged", updateLocalTimestamp) }
+
```
heliocliu, line 64:
```suggestion
```

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,75 @@
...
+      // TODO 4: Set the value of the localTimestamp state object that will appear in the UI.
+      const { sharedTimestamp } = this.fluidSharedObjects;
+      const updateLocalTimestamp = () => this.localTimestamp = { time: sharedTimestamp.get("time") };
+      updateLocalTimestamp();
+
+      // TODO 5: Register handlers.
+      sharedTimestamp.on("valueChanged", updateLocalTimestamp);
+
+      // TODO 6: Delete handler registration when the React App component is dismounted.
+      return () => { sharedTimestamp.off("valueChanged", updateLocalTimestamp) }
```
heliocliu, line 62: does angular use the same dismount logic as react?  definitely need to update the comment here, then adapt this section to match the typical angular lifecycle management if that's even relevant (looks like it might be since this return value is unused)

fluid-angular-starter/src/styles.css
```diff
@@ -0,0 +1 @@
+/* You can add global styles to this file, and also import other style files */
```
heliocliu, line 1: does angular still need this file even when it's empty?  otherwise probably worth removing here as well

scottn12: Yeah I actually tried to remove it but the angular compiler complained so I believe it is required

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,75 @@
...
+interface FluidDataModel { sharedTimestamp: SharedMap };
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+  styleUrls: ['./app.component.css']
+})
+export class AppComponent implements OnInit, OnDestroy {
+
+  fluidSharedObjects: FluidDataModel | undefined;
```
heliocliu, line 20: i think at this point it would be cleaner to get/save the sharedTimestamp directly instead of the whole sharedObjects, then extracting and saving the sharedTimestamp separately

fluid-angular-starter/src/app/app.component.ts
```diff
@@ -0,0 +1,75 @@
...
+  localTimestamp: TimestampDataModel | undefined;
+  sharedTimestamp: FluidDataModel['sharedTimestamp'] | undefined;
+  updateLocalTimestamp: ((...args: any[]) => void) | undefined;
```
heliocliu, line 23: this doesn't actually want any args so it would be better suited as a `() => void` or whatever the syntax is

heliocliu (approved):

fluid-angular-starter/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Modify the schema to include additional DDSes
+- Update the `defaultData` of those DDSes
+- Update the `model` to access and modify your Fluid data
+- Write custom events
+
+### Modifying the view
+
+- Modify the store
+  - `initialState`
```
micahgodbolt, line 29: this is all from the react readme and should be removed

micahgodbolt (approved):
Other than updating the readme to remove the old React starter, this is good. I'd still call consider calling this a 'demo' rather than a 'starter'.

heliocliu:
> Other than updating the readme to remove the old React starter, this is good. I'd still call consider calling this a 'demo' rather than a 'starter'.

This is a good point, probably worth doing the rename in this pr instead of later to something like `angular-demo`.  The topic here (syncing a timestamp) is what `cra-demo` is doing rather than `fluid-react-starter`, so that would create better parallelism with the examples Micah already wrote

skylerjokiel:
@scottn12 let's go with `@fluid-example/angular-demo` for this package. We have an item to update the other packages to reflect it.

fluid-angular-demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+Now, with our local service running in the background, we need to connect the application to it. The app has already been configured to this so now we just need to run the following in a new terminal window to start the app.
+
+```bash
+npm i
+npm run start
+```
+
+## Modifying the model
+
+### Specify additional DDSes
```
heliocliu, line 45: is the plan to add back the angular equivalent of what was here before in the followup?

heliocliu (approved):

fluid-angular-demo/README.md
```diff
@@ -0,0 +1,60 @@
+# fluid-angular-starter
```
skylerjokiel, line 1:
```suggestion
# @fluidframework/angular-demo
```

skylerjokiel (commented):
One more nit. Can you update the folder to `angular-demo` from `fluid-angular-demo`

skylerjokiel (approved):
🍁

skylerjokiel (approved):
🍁

PR #301, opened 2021-09-16
add instructions in view files for rendering that view

micahgodbolt:
I'm going to point directly to the view files in the docs, so I want to make sure that each view is able to define how to render that view.

### DIFF
--- a/multi-framework-diceroller/src/view/wcView.js
+++ b/multi-framework-diceroller/src/view/wcView.js
@@ -5,6 +5,19 @@
 
 import { diceValueKey } from "../app";
 
+// To see this view rendered, change the import in app.js to:
+// import { wcDiceRoller as diceRoller } from "./view";
+export const wcDiceRoller = (diceMap, elem) => {
+  customElements.define("wc-dice", Dice);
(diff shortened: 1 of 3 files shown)


multi-framework-diceroller/src/view/wcView.js
```diff
@@ -5,6 +5,19 @@
...
+  customElements.define("wc-dice", Dice);
+  const dice = document.createElement("wc-dice");
+  dice.onRoll = number => diceMap.set(diceValueKey, number);
+  const updateDice = () => dice.setAttribute("value", diceMap.get(diceValueKey));
+  updateDice();
+  diceMap.on("valueChanged", updateDice)
+
+  elem.append(dice);
+}
+
```
timtwang, line 20: Is there a specific reason for reorganizing where the `wcDiceRoller` is defined? Is it just for clearer clarity/readability?

micahgodbolt: yeah, I did it for consistency. All of the other view files had their primary rendering function at the top. If I moved the function down, and moved the comment down as well, i feel the comment would be buried.

PR #309, opened 2021-09-20
Address follow-up items for the Brainstorm audience demonstration

timtwang:
This is a continuation of the prior audience demonstration PR, aiming to address some of the follow-up items mentioned [here](https://github.com/microsoft/FluidFramework/issues/7303). This PR aims to:

1. Address the verbosity of the Brainstorm README
2. Add more comments in the code to explain audience more explicitly
3. Remove SharedMap object comparison
4. Add warning about object storing in SharedMap doesn't get stored the same way as it would be in a regular map

Some items that are mentioned in the issue but not addressed here are:

1. Extract styles into an external stylesheet
  - The current style used in Brainstorm is quite minimal and adding a stylesheet is actually more work. We can refactor the styles to a dedicated place when it gets more complex in the future.
2. Refactor Brainstorm view to allow key-based state update instead of regenerating the entire state each time
  - Given the scope of this item, it is likely that this will need some discussions for the redesign. We can address this at a later date.


brainstorm/README.md
```diff
@@ -41,7 +41,7 @@ This package is based on the [Create React App](https://reactjs.org/docs/create-
 
 1. Navigate to `http://localhost:3000`
 
-You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' the path `##1621961220840` is specifies one brainstorm document.
+  You'll be taken to a url similar to `http://localhost:3000/**#1621961220840**` the path `##1621961220840` specifies one brainstorm document.
 
 2. Create another chrome tab with `http://localhost:3000/**#1621961220840**`
```
heliocliu, line 46:
```suggestion
2. Navigate to the same url in another window or tab
```
Not really related to why this change is happening, but I'm not a fan of the existing wording

brainstorm/README.md
```diff
@@ -41,7 +41,7 @@ This package is based on the [Create React App](https://reactjs.org/docs/create-
 
 1. Navigate to `http://localhost:3000`
 
-You'll be taken to a url similar to 'http://localhost:3000/**#1621961220840**' the path `##1621961220840` is specifies one brainstorm document.
+  You'll be taken to a url similar to `http://localhost:3000/**#1621961220840**` the path `##1621961220840` specifies one brainstorm document.
```
micahgodbolt, line 44:
```suggestion
  You'll be taken to a url similar to `http://localhost:3000/**#1621961220840**` the path `#1621961220840` specifies one brainstorm document.
```

brainstorm/README.md
```diff
@@ -167,6 +166,14 @@ sharedMap.set(c_AuthorPrefix + noteId, newCardData.author);
 ```
 As shown above, a static prefix is attached to indicate which property this entry holds (`noteId`, `author`, `color`, etc), then to ensure the key is unique for each note, we attach the `noteId` after the static prefix. With this structure, we now ensured that properties for each note are stored individually.
 
+<br />
+
+| :warning: WARNING                                                                                                                                       |
+|:--------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `SharedMap` does not store object in the DDS map the same way it would be stored in a conventional map data structure. It is recommended to only store the necessary primitive data types in `SharedMap`.
```
heliocliu, line 173: I don't think advising to only store primitive types is quite right - the problem is more that comparison isn't valid for other types.  Maybe something like this would be better?
```suggestion
| `SharedMap` does not preserve object references like a conventional map data structure, and object comparisons of `SharedMap` values will be invalid . In this case, it is recommended to only store the necessary primitive data types in `SharedMap` or implement a custom comparison function.
```

brainstorm/README.md
```diff
@@ -210,88 +217,13 @@ The audience object also has a `getMyself()` function that returns the current c
...
 ```
 
-With `members` and `authorInfo` defined, we can use these to achieve several tasks:
+With `members` and `authorInfo` defined, we use these to achieve several tasks:
 
 1. displaying all current active users
-2. displaying author name in persona tooltip 
+    * All current active users are displayed as FacePile, or a list of personas, on the top right corner of the app.   
+2. displaying author name in persona tooltip
+    * When hovered over the note's persona, the author who created the name will be displayed dynamically.
```
heliocliu, line 225:
```suggestion
    * When hovering over the note's persona, the author who created the name will be displayed dynamically.
```
nit: matches tense of the parent point better

brainstorm/README.md
```diff
@@ -210,88 +217,13 @@ The audience object also has a `getMyself()` function that returns the current c
...
 3. displaying likes for the note
+    * When hovered over the like button, a list of all the users that liked the note will be shown.
```
heliocliu, line 227:
```suggestion
    * When hovering over the like button, a list of all the users that liked the note will be shown.
```
same nit

brainstorm/src/BrainstormModel.ts
```diff
@@ -25,8 +25,11 @@ export type BrainstormModel = Readonly<{
 }>;
 
 export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
+  // Global sharedMap that stores attributes of all the notes.
+  // The sharedMap can be updated by any user connected to the session
   const sharedMap: ISharedMap = fluid.initialObjects.map as SharedMap;
 
+  // check is note is complete
```
heliocliu, line 32:
```suggestion
  // check if note is complete
```

brainstorm/src/BrainstormModel.ts
```diff
@@ -37,13 +40,18 @@ export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
     return true;
   };
 
-
+  // Check if ote is deleted (noteId is 0)
```
heliocliu, line 43:
```suggestion
  // Check if note is deleted (noteId is 0)
```

brainstorm/src/BrainstormModel.ts
```diff
@@ -103,17 +115,23 @@ export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
 
     SetNoteColor,
 
+    // Set the note as liked or unliked by the author
```
heliocliu, line 118:
```suggestion
    // Set or unset the note as liked by the user
```
`author` was not the best variable name here :/

brainstorm/src/BrainstormModel.ts
```diff
@@ -103,17 +115,23 @@ export function createBrainstormModel(fluid: FluidContainer): BrainstormModel {
...
     LikeNote(noteId: string, author: AzureMember) {
       const voteString = c_votePrefix + noteId + "_" + author.userId;
-      sharedMap.get(voteString) === author
+
+    // WARNING: sharedMap does not store object in the DDS map the same way it would be stored in a conventional map data structure.
+    // Hence, instead of storing the entire AzureMember object, we are only storing the necessary primitive data types metadata.
+      sharedMap.get(voteString)?.userId === author.userId
         ? sharedMap.set(voteString, undefined)
-        : sharedMap.set(voteString, author);
+        : sharedMap.set(voteString, { userId: author.userId, userName: author.userName } );
```
heliocliu, line 126: does the retrieval side of this also need to be updated?  i'd guess it's fine at runtime because this inline object matches the used params, but if it's being typed/retrieved as an AzureMember that's not a good example of our own guidance

brainstorm/src/BrainstormModel.ts
timtwang: hmm but the only person who can trigger this method is the author 🤔

heliocliu: wait really?  you can't like other people's notes?

timtwang: my bad, I was thinking something else. Yes, `user` would be a better name here, and you can like other people's not since 
 `noteId` is an argument.

brainstorm/src/BrainstormModel.ts
timtwang: I don't think the value retrieved from the `sharedMap` is being typed as `AzureMember`. I was able to change the inline object params name and still have like feature working.

heliocliu (approved):

PR #317, opened 2021-09-27
Remove azureclient dep from mfwd

sumedhb1995:
This example doesn't use `AzureClient` so we can remove the package from its dependencies


micahgodbolt:
i did a double take on the time you made this PR, then realized you must be acclimating to your new time zone :D

micahgodbolt (approved):

PR #319, opened 2021-09-28
Use azure-local-service for brainstorm in local mode

sumedhb1995:


timtwang (approved):
thanks for updating the doc 😀

timtwang (approved):

PR #330, opened 2021-10-06
fixed outdated react-demo README

timtwang:
A user reached out to point out the outdated README for react-demo. This PR aims to address the [issue](https://github.com/microsoft/FluidFramework/issues/7726) the user created. 

We were not able to replicate the 2nd issue the user pointed out, so the implementation will remained as `location.hash` instead of `window.location.hash`.


heliocliu (approved):

timtwang:
> Does the readme code originate from some other source file? If so we should make sure to update that as well

The code originated from `App.js` but the code there is up-to-date, I think we just forgot to update the README after we made some changes.

PR #340, opened 2021-10-27
Add Collaborative Text Area example using SharedString

scottn12:
This PR adds a new example repo for using the `SharedString` DDS to create a collaborative text area in React.

Uses previously created example: https://github.com/microsoft/FluidFramework/blob/main/experimental/framework/react-inputs/src/CollaborativeTextArea.tsx


react-text-area/README.md
```diff
@@ -0,0 +1,39 @@
+# @fluid-example/react-text-area
+
+## About this repo
+
+This is an example to demonstrate building a collaborative Text Area in React using the `SharedString`. To use this example, we make the following assumptions:
+
+1. You want to use React.js for your view
+1. You already have Node installed on your local machine. If not, follow the instructions [here](https://nodejs.org/en/download/).
+
+For a more detailed explanation of this example please click [here](https://fluidframework.com/docs/recipes/shared-string/).
```
heliocliu, line 10: now that this example has been reframed as a text area guide, is it still going to be referenced as a shared string example on the ff site?

react-text-area/src/SharedStringHelper.js
```diff
@@ -0,0 +1,119 @@
...
+   * @returns The full text stored in the SharedString as a string.
+   */
+  getText() {
+    return this._latestText;
+  }
+
+  /**
+   * Insert the string provided at the given position.
+   */
+  insertText(text, pos) {
```
heliocliu, line 32: is this file taken from somewhere else?  if so would it be better to import it instead?

shared-string-example/src/CollaborativeTextArea.js
```diff
@@ -0,0 +1,136 @@
...
+   */
+  const handleChange = (ev) => {
+    // First get and stash the new textarea state
+    if (!textareaRef.current) {
+      throw new Error("Handling change without current textarea ref?");
+    }
+    const textareaElement = textareaRef.current;
+    const newText = textareaElement.value;
+    // After a change to the textarea content we assume the selection is gone (just a caret)
+    // This is a bad assumption (e.g. performing undo will select the re-added content).
```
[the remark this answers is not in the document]
heliocliu: It sounds like the omissions aren't going to be handled here (nor immediately after), so i think it makes sense to keep the comments for anyone who finds it and make sure we have an issue to track updating it later

react-text-area/src/SharedStringHelper.js
scottn12: This file was added in Matt's recent changes to the `CollaborativeTextArea` example in the main repo: https://github.com/microsoft/FluidFramework/blob/main/experimental/framework/react-inputs/src/SharedStringHelper.ts

skylerjokiel (approved):
Looks good. 👍

skylerjokiel (changes requested):
I totally didn't catch this before but just realized this is javascript and not typescript. 

Can you please port this example to typescript before committing.

heliocliu:
> I totally didn't catch this before but just realized this is javascript and not typescript.
> 
> Can you please port this example to typescript before committing.

This example is getting the walkthrough tutorial on the website as well, so i think js is right here

scottn12:
> I totally didn't catch this before but just realized this is javascript and not typescript.
> 
> Can you please port this example to typescript before committing.

I kept it in javascript since that is the default for `create-react-app`. Should I still port it to ts?

heliocliu (dismissed):

collaborative-text-area/src/SharedStringHelper.tsx
```diff
@@ -0,0 +1,146 @@
...
+    this._latestText = this._sharedString.getText();
+    const isLocal = event.isLocal;
+
+    const op = event.opArgs.op;
+    let transformPosition: (oldPosition: number) => number;
+
+    // Here we use `0` to represent an INSERT operation.
+    // Replace with `MergeTreeDeltaType.INSERT` when enum access issue is resolved.
+    // https://github.com/microsoft/FluidFramework/issues/8142
+    if (op.type === 0) {
```
scottn12, line 88: @anthony-murphy I had issues using the `MergeTreeDeltaType` enums and added this workaround for now. Does this seem okay for now?

collaborative-text-area/src/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+    else {
+      container = (await client.getContainer(containerId, containerSchema)).container;
+      if (!container.connected) {
+        await new Promise<void>((resolve) => {
+          container.once("connected", () => {
+            resolve();
+          });
+        });
+      }
```
scottn12, line 38: I noticed some issues with the container not actually being connected on app startup when getting an existing container, so I added this to ensure the container is connected before rendering the UI.

heliocliu: is there a specific scenario where this happens? e.g. only when navigating from a page that had a different instance of the example?

heliocliu: also that this is pretty unintuitive and should probably be fixed separately if it's truly intermittent

heliocliu (approved):

collaborative-text-area/src/App.tsx
scottn12: I find it is actually more common for `container.connected` to `false` than to be `true`. To reproduce I simply will load an instance of the example with an already existing container id.

scottn12: @heliocliu Do you think I should add a comment to this code block that this should be removed eventually?

heliocliu (approved):

PR #344, opened 2021-11-01
Added v1 of Fluid + Teams Application

timtwang:
In effort of the recipe that integrates Fluid into Teams, we are building a version 1 of an example application that will walk the developers through how the integration works. The goal of this PR are the following:

1. Create fluid container in the Config page and append the container ID onto the content page URL.
2. Developer can name the tab in the Config page.
3. Get and load the container in the content page.
4. Extract all container related operations into  `Util.ts`.
5. Convert all view and `SharedMap` update as a React component (`FluidContent`). 

The only files edited are `package.json`, `helloWorldTabConfig.tsx`, `helloWorldTab.tsx`, `Util.ts`, `FluidContent.tsx`. All other files are generated by Teams default template

TeamsTabHelloWorld

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -19,6 +19,9 @@ parameters:
 - name: multi-framework-diceroller
   type: string
   default: multi-framework-diceroller
+- name: node-demo
+  type: string
+  default: node-demo
 
 trigger:
 - main
--- a/node-demo/.gitignore
+++ b/node-demo/.gitignore
@@ -1,0 +1,4 @@
+dist
+node_modules
+*.log
+**/*.tsbuildinfo
--- a/node-demo/README.md
+++ b/node-demo/README.md
@@ -1,0 +1,19 @@
+# @fluid-example/node-demo
+
+This repository contains a simple NodeJS application connected with the Fluid Framework. Connected clients generate random numbers and display the result of any changes to the shared state.
+
+## Getting started
+
+Run Tinylicious server in the backgroud:
+
+```sh
+npm run start:server
+```
+
+Open a new terminal and run the client:
+
+```sh
+npm run start:client
+```
+
+To create a new Fluid container press Enter. The container id will be printed in the terminal. Copy the container id, launch a new terminal window, and type/paste the initial container id to have multiple collaborative NodeJS clients.
--- a/node-demo/package-lock.json
+++ b/node-demo/package-lock.json
@@ -1,0 +1,834 @@
+{
+  "name": "@fluid-example/node-demo",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@fluidframework/aqueduct": {
+      "version": "0.49.2",
+      "resolved": "https://registry.npmjs.org/@fluidframework/aqueduct/-/aqueduct-0.49.2.tgz",
+      "integrity": "sha512-gMMGzJue/OcJnDmBPwHx2JBVIjGQyUrRqOnuByeWQr727yWPgb9C2HC8+MkCRHrunOjOJMSAu1gUhslI6TFiqA==",
+      "requires": {
+        "@fluidframework/common-definitions": "^0.20.1",
+        "@fluidframework/common-utils": "^0.32.1",
+        "@fluidframework/container-definitions": "^0.40.0",
+        "@fluidframework/container-loader": "^0.49.2",
+        "@fluidframework/container-runtime": "^0.49.2",
+        "@fluidframework/container-runtime-definitions": "^0.49.2",
+        "@fluidframework/core-interfaces": "^0.39.7",
+        "@fluidframework/datastore": "^0.49.2",
+        "@fluidframework/datastore-definitions": "^0.49.2",
+        "@fluidframework/map": "^0.49.2",
+        "@fluidframework/request-handler": "^0.49.2",
+        "@fluidframework/runtime-definitions": "^0.49.2",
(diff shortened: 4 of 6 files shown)


teams-fluid-hello-world/package.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "teamsfluidhelloworld",
```
sumedhb1995, line 2:
```suggestion
  "name": "teams-fluid-hello-world",
```

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+            updateDice();
+            // Use the changed event to trigger the rerender whenever the value changes.
+            props.fluidMap.on("valueChanged", updateDice);
+
+            // turn off listener when component is unmounted
+            return () => { props.fluidMap.off("valueChanged", updateDice); };
+        }
+    }, [props.fluidMap]);
+
+    // Set the value at our dataKey with a random number between 1 and 6.
```
sumedhb1995, line 36:
```suggestion
    // Set the value at our diceValueKey with a random number between 1 and 6.
```

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+export type FluidContentProps = Readonly<{
+    fluidMap: SharedMap;
+}>
+
+export const FluidContent = (props : FluidContentProps) => {
+
+    const [diceView, setDiceView] = useState<{color : string, content : string} | undefined >(undefined);
+
+    useEffect(() => {
+        if (props.fluidMap !== undefined) {
```
sumedhb1995, line 16: This undefined check shouldn't be required since the prop interface doesn't accept undefined

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+            props.fluidMap.on("valueChanged", updateDice);
+
+            // turn off listener when component is unmounted
+            return () => { props.fluidMap.off("valueChanged", updateDice); };
+        }
+    }, [props.fluidMap]);
+
+    // Set the value at our dataKey with a random number between 1 and 6.
+    const onClick = () => {
+        if (props.fluidMap !== undefined) {
```
sumedhb1995, line 38: This undefined check shouldn't be required since the prop interface doesn't accept undefined

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,52 @@
...
+                getFluidMap(containerId);
+            });
+        } else {
+            const url = new URLSearchParams(window.location.search);
+            const containerId = url.get("containerId");
+            getFluidMap(containerId);
+        }
+    }, [inTeams]);
+
+    return (
```
sumedhb1995, line 49: Since `fluidMap` can be undefined, I think we should do a check here and that would also avoid the type casting
i.e.
```
if (fluidMap === undefined) {
   return <Loader label="Loading..." />;
} else {
   return <FluidContent fluidMap={fluidMap} />;
}
```

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const [{ inTeams, theme, context }] = useTeams({});
+    const [text, setText] = useState<string>();
+    const tabName = useRef("");
+
+    const onSaveHandler = async (saveEvent: microsoftTeams.settings.SaveEvent) => {
+        const host = "https://" + window.location.host;
+        createContainer().then((containerId) => {
+            microsoftTeams.settings.setSettings({
+                contentUrl: host + "/helloWorldTab/?containerId=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
```
sumedhb1995, line 21: Consider having `"containerId"` be a constant defined in Util.ts and then referenced here and in the content page

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,36 @@
...
+import { InsecureTokenProvider } from "@fluidframework/test-client-utils";
+
+export const diceValueKey = "dice-value-key";
+const containerSchema = {
+    initialObjects: { diceMap: SharedMap }
+};
+const connectionConfig : AzureClientProps =
+{
+    connection: {
+        tenantId: "8e6f1a42-e614-4077-99f3-d39d4a7ca140",
```
sumedhb1995, line 12: Uh-oh, you should avoid checking in the tenant ID and key of an FRS instance since it will always be in the git history. I would delete this instance immediately, and, at the least, revert back to an earlier commit that doesn't have this change and force-push this branch. Otherwise, ppl can start using this tenant

sumedhb1995 (changes requested):

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
sumedhb1995: Also, so that 3P developers also avoid a similar pitfall, consider seeing if we can fetch it from environment variables that aren't checked in instead

PR #345, opened 2021-11-01
Added v1 for Fluid + Teams application

timtwang:
In effort of the recipe that integrates Fluid into Teams, we are building a version 1 of an example application that will walk the developers through how the integration works. The goal of this PR are the following:

1. Create fluid container in the Config page and append the container ID onto the content page URL.
2. Developer can name the tab in the Config page.
3. Get and load the container in the content page.
4. Extract all container related operations into  `Util.ts`.
5. Convert all view and `SharedMap` update as a React component (`FluidContent`). 

The only files edited are `package.json`, `helloWorldTabConfig.tsx`, `helloWorldTab.tsx`, `Util.ts`, `FluidContent.tsx`. All other files are generated by Teams default template

TeamsTabHelloWorld


sumedhb1995 (commented):
Awesome work here Tim! This is shaping up really nicely 😄 I had left some comments on the earlier PR you had posted, would be great to see the same addressed here. Also, just for clarification, the README containing the recipe allowing developers to build this out for themselves starting from the teams boilerplate will be coming in a follow-up, correct?

heliocliu:
are the compile error flashes in the example video resolved?

timtwang:
> are the compile error flashes in the example video resolved?

yes it is! Updating `webpack.config.js` fixed it

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { diceValueKey } from "./Util";
+import { useState, useEffect } from "react";
+
+export type FluidContentProps = Readonly<{
+    fluidMap: SharedMap;
+}>
+
+export const FluidContent = (props : FluidContentProps) => {
+
+    const [diceView, setDiceView] = useState<{color : string, content : string} | undefined >(undefined);
```
skylerjokiel, line 13: Since you're passing in the SharedMap here you can get strong typing on this object.

```suggestion
    const [diceView, setDiceView] = useState<{color : string, content : string}>(props.fluidMap.get(diceValueKey));
```

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+            });
+        };
+        // ensure update runs at least once
+        updateDice();
+        // Use the changed event to trigger the rerender whenever the value changes.
+        props.fluidMap.on("valueChanged", updateDice);
+
+        // turn off listener when component is unmounted
+        return () => { props.fluidMap.off("valueChanged", updateDice); };
+    }, [props.fluidMap]);
```
skylerjokiel, line 32: Since you're passing in the SharedMap here you can get strong typing on this object.

I'm coding this in the browser without validation so please validate.

```suggestion
    const generateCurrentViewState = (): {color : string, content : string} => {
           const diceValue = props.fluidMap.get(diceValueKey);
            // Unicode 0x2680-0x2685 are the sides of a dice (⚀⚁⚂⚃⚄⚅)
            return {
                content: String.fromCodePoint(0x267f + diceValue),
                color: `hsl(${diceValue * 60}, 70%, 30%)`
            };
    }
    const [diceView, setDiceView] = useState<{color : string, content : string}>(generateCurrentViewState());

    useEffect(() => {
        // sync Fluid data into view state
        const updateDice = () => {
            setDiceView(generateCurrentViewState());
        };
        // Use the changed event to trigger the rerender whenever the value changes.
        props.fluidMap.on("valueChanged", updateDice);

        // turn off listener when component is unmounted
        return () => { props.fluidMap.off("valueChanged", updateDice); };
    }, [props.fluidMap]);
```

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from "react";
+import { SharedMap } from "fluid-framework";
+import { useState, useEffect } from "react";
```
skylerjokiel, line 3:
```suggestion
import { SharedMap } from "fluid-framework";
import * as React, { useState, useEffect } from "react";
```

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FluidContent } from "./FluidContent";
+import { getContainer, containerIdString } from "./Util";
+
+/**
+ * Implementation of the HelloWorldTab content page
+ */
+export const HelloWorldTab = () => {
+
+    microsoftTeams.initialize();
+    const [{ inTeams, theme, context }] = useTeams();
```
skylerjokiel, line 15: You're not using theme or context. Did these get accidentally removed or do you not need to use them?

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,41 @@
+import { SharedMap, IFluidContainer } from "fluid-framework";
+import { AzureClient, AzureClientProps, LOCAL_MODE_TENANT_ID } from "@fluidframework/azure-client";
+import { InsecureTokenProvider } from "@fluidframework/test-client-utils";
+import * as process from "process";
+import * as dotenv from "dotenv-webpack";
+
+export const diceValueKey = "dice-value-key";
+export const containerIdString = "containerId";
```
skylerjokiel, line 8: nit:

```suggestion
export const containerIdQueryParamKey = "containerId";
```

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,41 @@
...
+dotenv.config();
```
skylerjokiel, line 9: What's this doing?

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+import * as microsoftTeams from "@microsoft/teams-js";
+import { FluidContent } from "./FluidContent";
+import { getContainer, containerIdString } from "./Util";
+
+/**
+ * Implementation of the HelloWorldTab content page
+ */
+export const HelloWorldTab = () => {
+
+    microsoftTeams.initialize();
```
skylerjokiel, line 14: @timtwang , is this where the initialize happens in the generated code? This can get run multiple times depending on how the function renders. Just want to make sure that calling into this multiple times is okay. @sumedhb1995 , you know by chance?

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const [{ inTeams, theme, context }] = useTeams();
+
+    const [fluidMap, setFluidMap] = useState<SharedMap | undefined>(undefined);
+
+    useEffect(() => {
+        if (inTeams === true) {
+            microsoftTeams.appInitialization.notifySuccess();
```
skylerjokiel, line 21: What's the purpose of notifying success outside of the TabConfig?

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+    useEffect(() => {
+        if (inTeams === true) {
+            microsoftTeams.settings.getSettings(async (instanceSettings) => {
+                const url = new URLSearchParams(instanceSettings.contentUrl);
+                getFluidMap(url);
+            });
+        } else {
+            const url = new URLSearchParams(window.location.search);
+            getFluidMap(url);
+        }
```
skylerjokiel, line 45: I would say we should just not support non-teams scenarios. This will make the code clearer.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+        if (!containerId) {
+            throw Error("containerId not found in the URL");
+        }
+        getContainer(containerId).then((container) => {
+            const diceMap = container.initialObjects.diceMap as SharedMap;
+            setFluidMap(diceMap);
+        });
+    };
+
+    useEffect(() => {
```
skylerjokiel, line 36: We have two useEffects here that have the same conditional code. Let's consolidate.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const [fluidMap, setFluidMap] = useState<SharedMap | undefined>(undefined);
+
+    useEffect(() => {
+        if (inTeams === true) {
+            microsoftTeams.appInitialization.notifySuccess();
+        }
+    }, [inTeams]);
+
+    const getFluidMap = (url : URLSearchParams) => {
```
skylerjokiel, line 25: Move this within the context of the useEffect it's being used.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { createContainer, containerIdString } from "./Util";
+
+/**
+ * Implementation of HelloWorldTab configuration page
+ */
+export const HelloWorldTabConfig = () => {
+
+    const [{ inTeams, theme, context }] = useTeams({});
+    const [text, setText] = useState<string>();
+    const tabName = useRef("");
```
skylerjokiel, line 15: Is this tabName code part of the standard generation? If it is then feel free to leave it. If it's not then I wouldn't add it.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
```diff
@@ -0,0 +1,65 @@
...
+        createContainer().then((containerId) => {
+            microsoftTeams.settings.setSettings({
+                contentUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
+                websiteUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
+                suggestedDisplayName: tabName.current,
+                removeUrl: host + "/helloWorldTab/remove.html?theme={theme}",
+                entityId: tabName.current
+            });
+        }).catch(error => console.log(error));
+        saveEvent.notifySuccess();
```
skylerjokiel, line 28: We should only notifySuccess if the container is created successfully.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
```diff
@@ -0,0 +1,65 @@
...
+        const host = "https://" + window.location.host;
+        createContainer().then((containerId) => {
+            microsoftTeams.settings.setSettings({
+                contentUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
+                websiteUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
+                suggestedDisplayName: tabName.current,
+                removeUrl: host + "/helloWorldTab/remove.html?theme={theme}",
+                entityId: tabName.current
+            });
+        }).catch(error => console.log(error));
```
skylerjokiel, line 27: We're in an async function and you're calling an async call. Also it's bad practice to eat and console log errors that would cripple an application. Either handle them correctly or let them bubble up.

```suggestion
        const contaienrId = await createContainer();
        microsoftTeams.settings.setSettings({
            contentUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
            websiteUrl: host + "/helloWorldTab/?" + containerIdString + "=" + containerId + "&name={loginHint}&tenant={tid}&group={groupId}&theme={theme}",
            suggestedDisplayName: tabName.current,
            removeUrl: host + "/helloWorldTab/remove.html?theme={theme}",
            entityId: tabName.current
        });
```

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
skylerjokiel: This should just happen correctly by moving to the async logic above.

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,41 @@
...
+    initialObjects: { diceMap: SharedMap }
+};
+
+const connectionConfig : AzureClientProps =
+{
+    connection: {
+        tenantId: LOCAL_MODE_TENANT_ID,
+        // tokenProvider: new InsecureTokenProvider("foobar", { id: "user" }),
+        // orderer: "http://localhost:7070",
+        // storage: "http://localhost:7070"
```
skylerjokiel, line 20: Clean this up.

```suggestion
```

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,41 @@
...
+        tokenProvider: new InsecureTokenProvider(process.env.REACT_APP_TENANT_KEY as string, { id: "user" }),
+        orderer: process.env.REACT_APP_ORDERER as string,
+        storage: process.env.REACT_APP_STORAGE as string
```
skylerjokiel, line 23: I feel like the environment variables are overkill here. I feel like hard coding the values and adding a comment block (or description in the recipe) of how to add your own experience would provide a better learning experience.

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+};
+console.log("ab " + process.env.PACKAGE_NAME);
```
skylerjokiel, line 26: Lets remove unnecessary console logs.

```suggestion
```

teams-fluid-hello-world/src/server/server.ts
```diff
@@ -0,0 +1,68 @@
+import * as Express from "express";
```
skylerjokiel, line 1: Oh interesting! This comes with a server...

I could see a followup to this work would be writing a Token Provider that talks to this server.

skylerjokiel (changes requested):
Awesome work. I've added some comments mainly around code cleanup.

One thing that's hard for me to grok by reading this is what you've added vs what is in a native generated Teams Tab example. I think the best thing we can do here is minimize the number to deltas between the original and adding Fluid. If something I asked to be removed was in the original generated code lets just leave it. The less we have to explain the easier :)

timtwang:
> Awesome work here Tim! This is shaping up really nicely 😄 I had left some comments on the earlier PR you had posted, would be great to see the same addressed here. Also, just for clarification, the README containing the recipe allowing developers to build this out for themselves starting from the teams boilerplate will be coming in a follow-up, correct?

Thanks for taking a pass Sumedh. I have addressed all the comments you made including hiding the credentials in the .env file. And yes, the README will come as a follow-up. 😀

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    // Set the value at our diceValueKey with a random number between 1 and 6.
+    const onClick = () => {
+        props.fluidMap.set(diceValueKey, Math.floor(Math.random() * 6) + 1);
+    };
+
+    if (!diceView) {
+        return <div>Loading dice...</div>;
+    }
```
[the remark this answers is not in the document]
timtwang: right, there is a fluidMap check in `HelloWorldTab.tsx`. I will remove it

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
timtwang: I'm not using theme or context, so I'll remove them 👍

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
timtwang: yup. Took care of this since @sumedhb1995 and I were trying to hide the creds in an .env so we don't commit creds accidentally again

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
timtwang: I agree. I was a bit worried that hiding the credentials with environment variables would be too tedious and distracting for the developers and this is something they can look into if they want their app to be more secure.

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTabConfig.tsx
timtwang: make sense. It was actually preventing the entire config page from working if it's unable to create the client and create the container.

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
timtwang: works like a charm!! 💯

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
timtwang: Typescript is giving some compilation error for this as it expecting a string literal, so I'll be keeping the original version

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
sumedhb1995: @skylerjokiel Tim and I were debating this as well, and I would actually have the opposite opinion. I think the recipe should guide developers on what the best practices are. Having the tenant ID/key out in the open leaves it ripe for checking in accidentally. And I'm sure that a lot of developers who clone this code will accidentally do this. Once it is in the git history, its impossible to get it out and they would have to go nuke their FRS tenant to be properly protected for others misusing their tenant.

The Teams generated code already provides a .env file that it is using and this would only be adding on to that so I don't think it would be overkill or adding much additional overhead to it to have this check in. Also, figuring out how to pass the env variables does take some finagling with webpack config so we would be saving them time by giving them a solution to that problem instead of having to solve it on their own

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
sumedhb1995: It's okay for it to run multiple times. However, this needs to happen before any of the SDK functions are called and I'd recommend having it in the client.ts page instead so that it only runs once.
@timtwang is this what the generator provided? Its a bit strange since the SDK functions are also used in the config page and they shouldn't work since initialize hasn't been called prior to loading it.
https://docs.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/using-teams-client-sdk#basic-functions

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
sumedhb1995: There are two `.notifySuccess` calls in TabConfig. One of them is for the `saveEvent` which only needs to happen there. The `appInitialization` one needs to happen for all of them to let the Teams main app know the page successfully loaded

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
```diff
@@ -0,0 +1,34 @@
...
+import { AzureClient, AzureClientProps, LOCAL_MODE_TENANT_ID } from "@fluidframework/azure-client";
+import { InsecureTokenProvider } from "@fluidframework/test-client-utils";
+
+export const diceValueKey = "dice-value-key";
+export const containerIdQueryParamKey = "containerId";
+const containerSchema = {
+    initialObjects: { diceMap: SharedMap }
+};
+
+const connectionConfig : AzureClientProps =
```
sumedhb1995, line 11: I'm of the opinion that for a Teams Tab example, we really should have a non-local, FRS based config example alongside the local one. For hello world examples, it makes sense to keep it as just local. But for a Teams Tab, where the focus of having it in a tab experience in the first place is to share it with other ppl in a channel/meeting, it doesn't provide much in the way of usability if noone outside the current computer can use it

teams-fluid-hello-world/src/client/helloWorldTab/FluidContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { diceValueKey } from "./Util";
+import { useState, useEffect } from "react";
+
+export type FluidContentProps = Readonly<{
+    fluidMap: SharedMap;
+}>
+
+export const FluidContent = (props : FluidContentProps) => {
+
+    const generateCurrentViewState = (): {color : string, content : string} => {
```
sumedhb1995, line 13: nit verbosity since `state` in a React context means that it is for the current view:
```suggestion
    const generateState = (): {color : string, content : string} => {
```

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
sumedhb1995: Agreed, we can have `microsoftTeams.appInitialization.notifySuccess();` be chained to the `microsoftTeams.settings.getSettings` call here

sumedhb1995 (commented):
Nice! Thanks for integrating all the edits in from our call! 😄 The only main thing that is a stickler for me is that I think we need to include a config example for connecting with an FRS tenant using .env variables. @skylerjokiel @timtwang happy to chat about this!

teams-fluid-hello-world/src/client/helloWorldTab/HelloWorldTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+    useEffect(() => {
+        if (inTeams === true) {
+            microsoftTeams.settings.getSettings(async (instanceSettings) => {
+                const url = new URL(instanceSettings.contentUrl);
+                getFluidMap(url.searchParams);
+            });
+        } else if (inTeams === false) {
+            const url = new URL(window.location.search);
+            getFluidMap(url.searchParams);
+        }
```
skylerjokiel, line 44: I don't think we need to handle the non-Teams case. I would just say we should put up a div that say "This application only works in the context of Teams" or something like that.

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
skylerjokiel: 100% agree @sumedhb1995, but I don't think we need two examples. I think this example should have local as default because it means we can run local testing. In the recipe we should have a section that talks about how to change these values.

I would say we should add some comments here like in HelloWorld that tells people to past their info.

skylerjokiel (dismissed):
PR looks good.

Please add this to the build process by updating the build.yml file at the root.

teams-fluid-hello-world/src/client/helloWorldTab/Util.ts
timtwang: I added comments such as `ENTER YOUR TENANT-ID HERE` just like HelloWorld

sumedhb1995 (approved):
Looks good! Let's discuss .env storing the azure tenant ID and key later on though. Not a blocker for this PR but I still think it would be good to have!
Nice work! 🚢

timtwang:
> Looks good! Let's discuss .env storing the azure tenant ID and key later on though. Not a blocker for this PR but I still think it would be good to have! Nice work! 🚢

yup! I currently have the .env approach mentioned in the walk-thru as a note but we can talk about it more

PR #353, opened 2021-11-19
Refactor BrainstormExample using react hooks instead of ViewModel des…

drazvan91:
I did full refactoring of BrainstormExample using react hooks instead of ViewModel design pattern and passing it top-down through components. In the ```core``` folder there is one important hook called `useFluidState(key)` which allows you to get, set and delete an entry in the SharedMap (which is exposed to the entire application using React Context).

It works similar to```useState()```. Here is a simple example of using it for a dummy NoteComponent:
```javascript
function MyNoteComponent(props: {noteId: string}){
    const { noteId } = props;
    const [ noteText, setNoteText, deleteNoteText ] = useFluidState<string>("noteText_" + noteId);

    const value = noteText || ''; // in case key does not exist, the noteText is undefined

    return <div>
       <input value={ value } onChange={ ev => setNoteText(ev.currentTarget.value) } />
       <button onClick={ deleteNoteText } > Reset </button>
       <button 
    </div>
}
```

It's worth to be mentioned that ```useFluidState``` listens only for the specified key and the component will not be rerendered if other keys from the SharedMap are changed.

Now, instead of having a BrainstormModel there is a set of hooks in ```brainstorm-hooks.ts``` which contains business logic for filtering, updating and managing notes.

Besides that I added support for MouseCursors, so now you can see other users moving their mouse cursors around.

The connection to the Container is also managed by a hook called ```use-fluid-connection.ts```. Here is how it works in `App.tsx`:

```javascript
export function App() {
  const { container, services, isLoading, error } = useFluidConnection(
    connectionConfig,
    containerSchema,
  );

  if (isLoading) {
    return <div>Connecting, please wait</div>;
  }
  if (error) {
    return <div>{error}</div>;
  }

  return (
    <ConnectedApp container={container!} services={services!}></ConnectedApp>
  );
}
```

What do you think?


drazvan91:
Hey @scottn12, what do you think of this pull request ?

brainstorm/src/contexts/fluid-context.ts
```diff
@@ -0,0 +1,11 @@
+import { AzureContainerServices } from '@fluidframework/azure-client';
+import { IFluidContainer } from '@fluidframework/fluid-static';
+import { ISharedMap } from 'fluid-framework';
+import React from 'react';
+
+export interface FluidContextType {
+  container: IFluidContainer;
+  services: AzureContainerServices;
+  sharedMap: ISharedMap;
```
scottn12, line 9: Could the type here be `SharedMap` instead of `ISharedMap`?

brainstorm/src/Themes.ts
```diff
@@ -0,0 +1,80 @@
...
+    neutralTertiaryAlt: '#000000',
+    neutralTertiary: '#c8c8c8',
+    neutralSecondary: '#d0d0d0',
+    neutralPrimaryAlt: '#dadada',
+    neutralPrimary: '#ffffff',
+    neutralDark: '#f4f4f4',
+    black: '#f8f8f8',
+    white: '#000000',
+  },
+});
```
scottn12, line 80: We should remove the code related to theming since as we want the example to focus on code related to using FluidFramework. Having additional code/files make it more difficult for new developers to find the relevant code.
```suggestion

```

brainstorm/src/Types.ts
```diff
@@ -1,22 +1,56 @@
...
+export const DefaultColor = ColorOrder[0];
+
+export const ColorOptions: { [key in ColorId]: ColorValues } = {
+  Blue: { base: '#0078D4', dark: '#99C9EE', light: '#CCE4F6' },
+  Green: { base: '#005E50', dark: '#99BFB9', light: '#CCDFDC' },
+  Yellow: { base: '#F2C811', dark: '#FAE9A0', light: '#FCF4CF' },
+  Pink: { base: '#E3008C', dark: '#F499D1', light: '#F9CCE8' },
+  Purple: { base: '#8764B8', dark: '#CFC1E3', light: '#E7E0F1' },
+  Orange: { base: '#CA5010', dark: '#EAB99F', light: '#F4DCCF' },
+};
```
scottn12, line 56: More theme related code
```suggestion
```

scottn12 (changes requested):
@drazvan91, this is very cool, thank you for contributing to our examples! I didn't get a chance to go over all of the code yet, but I left a few comments for what I have looked over.

I did notice one major issue while loading the application. When I open a new session of the application I got stuck on the "Logging you in. Please wait" message until I refreshed. Then I got stuck on the "Connecting, please wait" message. After another refresh it loaded properly and worked well. Would you be able to revisit the loading logic for this PR?

ChumpChief:
The brainstorm example has undergone substantial changes since this PR was open (moving to FF 2.0 and switching to using SharedTree).  Since there's not been recent activity, closing this PR.

PR #417, opened 2022-02-21
Replace readline-async with readline-sync

alexvy86:
`readline-async` seems to be causing `Ctrl-C` not to work once the application is running, so replacing it with a simpler option that keeps it working.

Partially fixes https://github.com/microsoft/FluidFramework/issues/9208 (pending a PR to fix the [corresponding code in the documentation](https://github.com/microsoft/FluidFramework/blob/main/docs/content/docs/recipes/node.md)).

### DIFF
--- a/node-demo/package.json
+++ b/node-demo/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "@fluidframework/tinylicious-client": "^0.49.2",
     "fluid-framework": "^0.49.2",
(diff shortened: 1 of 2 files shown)


tylerbutler:
Thanks, @alexvy86!

The CI failure is due to some known issues with the action config. <https://github.com/microsoft/FluidFramework/pull/9229> should address that, so I'm merging this change.

PR #418, opened 2022-02-23
Shared object name change in nodejs example

sdeshpande3:


tylerbutler (approved):
Sorry for the delay! :shipit:

PR #512, opened 2022-07-08
small code cleanup

RishhiB:
# Code cleanup
Small code cleanup in Fluid Examples by removing else statement by returning early, using hooks instead of classical components, and object destructuring container

### DIFF
--- a/collaborative-text-area/src/App.tsx
+++ b/collaborative-text-area/src/App.tsx
@@ -4,14 +4,14 @@
  */
 
 import "./App.css";
-import React from "react";
+import { useState, useEffect } from "react";
 import { TinyliciousClient } from "@fluidframework/tinylicious-client";
 import { ConnectionState, ContainerSchema, IFluidContainer, SharedString } from "fluid-framework";
 import { CollaborativeTextArea } from "./CollaborativeTextArea";
 import { SharedStringHelper } from "@fluid-experimental/react-inputs";
 
 const useSharedString = (): SharedString => {
-  const [sharedString, setSharedString] = React.useState<SharedString>();
+  const [sharedString, setSharedString] = useState<SharedString>();
   const getFluidData = async () => {
     // Configure the container.
     const client: TinyliciousClient = new TinyliciousClient();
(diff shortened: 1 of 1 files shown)


collaborative-text-area/src/App.tsx
```diff
@@ -23,26 +23,27 @@ const useSharedString = (): SharedString => {
     let container: IFluidContainer;
     const containerId = window.location.hash.substring(1);
     if (!containerId) {
-      container = (await client.createContainer(containerSchema)).container;
+      ({ container } = await client.createContainer(containerSchema));
```
ChumpChief, line 26:
```suggestion
      { container } = await client.createContainer(containerSchema);
```

collaborative-text-area/src/App.tsx
```diff
@@ -23,26 +23,27 @@ const useSharedString = (): SharedString => {
...
     }
-    else {
-      container = (await client.getContainer(containerId, containerSchema)).container;
-      if (container.connectionState !== ConnectionState.Connected) {
-        await new Promise<void>((resolve) => {
-          container.once("connected", () => {
-            resolve();
-          });
+
+    ({ container } = await client.getContainer(containerId, containerSchema));
```
ChumpChief, line 33:
```suggestion
    { container } = await client.getContainer(containerId, containerSchema);
```

Josmithr (approved):

collaborative-text-area/src/App.tsx
```diff
@@ -23,26 +23,27 @@ const useSharedString = (): SharedString => {
...
     const containerId = window.location.hash.substring(1);
     if (!containerId) {
-      container = (await client.createContainer(containerSchema)).container;
+      ({ container } = await client.createContainer(containerSchema));
       const id = await container.attach();
       window.location.hash = id;
+      // Return the Fluid SharedString object.
+      return container.initialObjects.sharedString as SharedString;
     }
-    else {
```
ChumpChief, line 30: Prob prefer to keep the if/else here rather than early exit, since we're choosing one of two distinct flows rather than partial vs. full flow.

collaborative-text-area/src/App.tsx
Josmithr: I think the parens are necessary since container is already declared, right?

ssimic2 (approved):

scottn12 (approved):

PR #543, opened 2022-10-15
Setup prettier in all examples

sonalivdeshpande:


Josmithr:
I would recommend using our baseline eslint and prettier configs from fluid-framework if possible.

- eslint: https://github.com/microsoft/FluidFramework/blob/main/common/build/eslint-config-fluid/recommended.js
- prettier: https://github.com/microsoft/FluidFramework/blob/main/common/build/build-common/prettier.config.cjs

Josmithr:
I would also recommend adding prettier as the default formatter in the vscode settings file, and adding "format-on-save". That way developer will get the desired formatting without having to run the script.

```json
"editor.defaultFormatter": "esbenp.prettier-vscode",
"editor.formatOnSave": true,
```

angular-demo/.prettierignore
```diff
@@ -0,0 +1,12 @@
...
+
+# Dependencies
+node_modules
+
+# Build output
+dist
+build
+
+# VSCode settings
+.vscode
```
Josmithr, line 12: Do we need to ignore the vscode settings? Prettier does a pretty good job formatting json.

Josmithr:
Is format on save working correctly in vscode with the current setup? Just wondering if we need to add a dependency on prettier from the root package.json + add the format-on-save config to the repo root vscode settings.

Josmithr (approved):
I would suggest extending the base prettier config from build-common in these packages as well. Otherwise, looks good to me!

Josmithr (approved):

Josmithr (approved):

sonalivdeshpande:
> Is format on save working correctly in vscode with the current setup? Just wondering if we need to add a dependency on prettier from the root package.json + add the format-on-save config to the repo root vscode settings.

setting the format-on-save on the repo root works. thanks for pointing that out 😄

PR #545, opened 2022-10-18
Configure dependabot to open prs for specific packages

sonalivdeshpande:


ChumpChief:
This is stale now since most of these examples have been removed.  Opened #1047 to do the same thing but with the current example set.

PR #585, opened 2022-11-01
Setup jest in examples

sonalivdeshpande:
Setting up jest infrastructure in the following examples:

1. react-demo
2. react-starter-template
3. audience-demo
4. collaborative text area
5. multi framework diceroller
6. brainstorm
7. angular-demo


brainstorm/.env
```diff
@@ -0,0 +1 @@
+BROWSER=none
```
tylerbutler, line 1: Generally .env files are not committed. Can this default be set in code instead?

collaborative-text-area/package.json
```diff
@@ -24,13 +24,18 @@
...
-		"@fluidframework/build-common": "^1.1.0"
+		"@fluidframework/build-common": "^1.1.0",
+		"jest": "^29.2.2",
+		"jest-puppeteer": "^6.1.1",
+		"puppeteer": "^19.2.1",
+		"start-server-and-test": "^1.14.0"
 	},
 	"scripts": {
 		"start": "react-scripts start",
+		"start:server": "npx tinylicious@latest",
```
tylerbutler, line 35: suggest adding tinylicious as a dev dependency for any project that uses it for testing. That way these commands use the local version rather than pulling latest from npm, which is usually slow.

collaborative-text-area/test/collaborativetext.test.ts
```diff
@@ -0,0 +1,46 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+const assert = require("console").assert;
```
tylerbutler, line 6:
```suggestion
import { strict as assert } from "assert";
```

collaborative-text-area/test/collaborativetext.test.ts
```diff
@@ -0,0 +1,46 @@
...
+const config = require("../jest.config");
+
+let url;
+
+describe("collaborative-text-area", () => {
+	const load = async () => {
+		await page.goto(config.globals.URL, {
+			waitUntil: ["networkidle2", "load"],
```
tylerbutler, line 14: Maybe a comment explaining why this is needed?

react-starter-template/test/App.test.tsx
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+const assert = require("console").assert;
```
tylerbutler, line 6:
```suggestion
import { strict as assert } from "assert";
```

tylerbutler (approved):

tylerbutler (approved):
Double-check the assert imports. VSCode often picks one for you when you try to use it, but not the right one.

tylerbutler (approved):

PR #596, opened 2022-11-09
Remove lerna

sonalivdeshpande:
Removing `lerna` and adding `syncpack` instead to maintain same version across the repository.


package.json
```diff
@@ -21,5 +12,11 @@
   "bugs": {
     "url": "https://github.com/microsoft/FluidFramework/issues"
   },
-  "homepage": "https://github.com/microsoft/FluidExamples#readme"
+  "homepage": "https://github.com/microsoft/FluidExamples#readme",
+  "dependencies": {
+    "syncpack": "^8.3.9"
+  },
+  "scripts": {
+    "fix-mismatches": "syncpack fix-mismatches --source */package.json"
```
tylerbutler, line 20: Could we add a test script that checks that the dependencies are all set correctly? That way the build will fail in PR when you don't upgrade uniformly.

tylerbutler (dismissed):

tylerbutler (approved):

PR #636, opened 2022-12-08
[TaskManager] Add TaskManager Example

scottn12:
## Description

This PR creates a new example to demonstrate the TaskManager DDS.

### Note: This PR will remain a draft until TaskManager is published. 
Once TaskManager is published this PR will be updated to use the newly published TaskManager version, and the README will be updated to include the URL of the corresponding recipe. 

## Reviewer Guidance

- General thoughts about the example's ability to demonstrate how to use TaskManager
- Is the code easy to follow/understand?
- Additional comments that could help clarify

## Misc

AB#2212


task-manager-diceroller/src/App.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+	React.useEffect(() => {
+		if (sharedMap !== undefined && taskManager !== undefined) {
+			// Update dice value from SharedMap
+			const syncView = () => setDiceValue(sharedMap.get(diceKey) ?? 1);
+			syncView();
+			sharedMap.on("valueChanged", syncView);
+
+			// TaskManager Setup
+			let rollInterval: NodeJS.Timeout | undefined;
```
ChumpChief, line 63: I think this is technically more correct (so it has correct typings in both Node and the browser):

```suggestion
			let rollInterval: ReturnType<typeof setInterval> | undefined;
```

task-manager-diceroller/src/App.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+			taskManager.on("assigned", startRollingDice);
+			taskManager.on("lost", stopRollingDice);
+			taskManager.on("completed", stopRollingDice);
+
+			// Once our listeners are setup we can subscribe to the task
+			taskManager.subscribeToTask(diceRollTaskId);
+
+			// Turn off listeners when component is unmounted
+			return () => {
```
ChumpChief, line 109: Nice cleanup!

ChumpChief (dismissed):
Few suggestions but looks good!

ChumpChief (approved):

task-manager-diceroller/README.md
```diff
@@ -0,0 +1,27 @@
...
+For a more detailed explanation of this example please click [here](TODO_RECIPE_LINK_HERE).
+
+## Using this repo
+
+### Run the app locally
+
+To run our local server, Tinylicious, on the default URL of `localhost:7070`, enter the following into a terminal window:
+
+```
+npx tinylicious
```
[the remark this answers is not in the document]
Josmithr: I noticed this was a mixed bag too. Would be nice to have consistency here.

Since these are examples to get users up and running with their own code `npx` might be nice, as it avoids obfuscating what they might do while prototyping their own code. But we could also document what `start:server` does instead. Consistency is my main concern.

task-manager-diceroller/README.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+Now, with our local service running in the background, we need to connect the application to it. The app has already been configured to this so now we just need to run the following in a new terminal window to start the app.
+
+```bash
+npm i
+npm run start
```
Josmithr, line 23: Nit
```suggestion
npm start
```

PR #731, opened 2023-02-19
De-registering function in React

Vishruta-Patil:
## Related Issue
[#11794 ](https://github.com/microsoft/FluidFramework/issues/11794)

## Description
The React Demo obtains a connected container with an useEffect hook like this:

```
const [fluidMap, setFluidMap] = React.useState(undefined);

React.useEffect(() => {
    getMyMap().then(myMap => setFluidMap(myMap));
}, []);

```
where getMyMap creates or connects to a new container and returns one of the shared DDS'.

However, this can result in multiple active container connections as described in https://github.com/microsoft/FluidFramework/issues/11304. The issue is that react can sometimes re-run the useEffect hook for various reasons, resulting in the creation of another active connection without cleaning up the previous one properly.

The solution  is to give react a de-registration function that cleans up the old connection (disconnect the container), when the useEffect function is re-run.


Vishruta-Patil:
@tylerbutler Can you plesae review the PR ?

Vishruta-Patil:
@microsoft-github-policy-service agree

tylerbutler:
@Vishruta-Patil Sorry for the delay! I'll review this soon!

Vishruta-Patil:
@tylerbutler Can I work on the documentation part on [this](https://github.com/microsoft/FluidFramework) till the time you review the PR,  as mentioned in this issue 
[11794](https://github.com/microsoft/FluidFramework/issues/11794)

react-demo/package-lock.json
tylerbutler, line 1: You shouldn't need to make any changes to this file, so please revert it.

react-demo/src/App.js
```diff
@@ -51,6 +51,20 @@ function App() {
...
+		  const myMap = container?.initialObjects?.myMap;
+		  setFluidMap(myMap)
+		})();
+	  
+		// Return a de-registration function that disconnects the old container whenever the hook re-runs
+		return () => {
+		  if (container != null) container?.disconnect();
+		};
+	  }, []);
+
```
tylerbutler, line 67: You shouldn't add a new call to useEffect here. Rather, all existing uses of useEffect (like the two above) need to be checked and if they don't return a dereg function, they should.

react-demo/src/App.js
```diff
@@ -51,6 +51,20 @@ function App() {
...
+		let container;
+		(async () => {
+		  container = await getContainer();
+		  const myMap = container?.initialObjects?.myMap;
+		  setFluidMap(myMap)
+		})();
+	  
+		// Return a de-registration function that disconnects the old container whenever the hook re-runs
+		return () => {
+		  if (container != null) container?.disconnect();
```
tylerbutler, line 64:
```suggestion
		  if (container != undefined) container?.disconnect();
```

I think you want to check `undefined` here, not `null`.

tylerbutler (changes requested):
@Vishruta-Patil Thank you for creating this PR! It's a great start. I think there are more useEffect calls that should be updated to return dereg functions. Please take a look at my comments and let me know if you have any questions. Thanks!

react-demo/src/App.js
rsommerfeld: Checking `container == null` returns `true` for both `undefined` and `null`, but checking `container === undefined` can be a bit more explicit in what we are testing for.

ChumpChief:
This example (react-demo) has since been removed from the repo, so closing this PR.

PR #762, opened 2023-04-18
Svelte + Vite example

RishhiB:
Simple Svelte Example, copied from the angular demo


svelte-demo/vite.config.ts
scottn12, line 1: Could you add the MSFT copyright comment to the typescript/javascript files

svelte-demo/src/main.ts
```diff
@@ -0,0 +1,8 @@
+import './app.css'
+import App from './App.svelte'
+
+const app = new App({
+  target: document.getElementById('app'),
+})
+
+export default app
```
scottn12, line 8:
```suggestion
import './app.css';
import App from './App.svelte';

const app = new App({
  target: document.getElementById('app'),
});

export default app;
```

svelte-demo/package.json
scottn12, line 1: I think we want to integrate prettier into this demo?

scottn12 (commented):
Looking good, thanks for starting this demo! One thing I didn't mention below is that we will probably want to add this demo to the `build.yml` file and potentially add some simple tests

ChumpChief:
Looks like this PR is stale - if you're interested in pursuing this example further, let's discuss.  If so, it would at least need to be updated to FF 2.0 and we'd want to discuss what DDSs should be used (i.e. should it use SharedTree).

PR #768, opened 2023-05-09
build: Misc. cleanup

Josmithr:
This PR includes misc. cleanup done while verifying all of the apps are working as intended. They include:

* package-lock updates
* Misc. documentation improvements (mostly formatting)
* Updating package app start flows to ensure all packages use the same flow (`npx tinylicious` in one terminal, `npm install` and `npm start` in another).
* Fix a potential bug in `BrainstormView` where React state was not correctly updated when dependency (container) changed. Fix is to use `useMemo` in place of `useState`.


sonalideshpandemsft (dismissed):
Looks good!! 💯 thanks for making all these improvements!!

sonalideshpandemsft (approved):

PR #769, opened 2023-05-25
Brainstorm demo fix

zhenmichael:
existing implementation does not pass in a name field for user configuration even though it is mandatory for remote Azure Connection. Replaced the id object with userConfig.


sonalideshpandemsft (approved):
:shipit:

PR #770, opened 2023-06-09
build: Update repo to Node version 16 to match the other Fluid repos

Josmithr:
Updates lockfiles to the npm v8 format, and updates some READMEs to clarify usage instructions.


alexvy86:
I don't think I can come up with a great answer myself, but do we want to keep these in sync with the development in main or with the latest GA version? I kind of lean towards the latter...

Josmithr:
> I don't think I can come up with a great answer myself, but do we want to keep these in sync with the development in main or with the latest GA version? I kind of lean towards the latter...

Hmm. That's an interesting point. Since no code changes are required, it may not matter too much in this case. But I definitely see your point. @ssimic2 what do you think?

scottn12:
> I don't think I can come up with a great answer myself, but do we want to keep these in sync with the development in main or with the latest GA version? I kind of lean towards the latter...

@alexvy86, I agree that we should keep the examples here on the latest GA release. IMO for a few reasons:
1. It is costly to update the examples every time we have a new internal release
2. It is a good exercise for our GA major/minor releases to update the examples here and observe any pain points
3. Since these examples are intended for users new to Fluid, it feels appropriate to use the GA releases

scottn12 (approved):
Thanks for doing this Josh :)

alexvy86:
> I think Alejandro's point was that our current GA release was developed on Node v14, rather than 16. So maybe we should not do this update until we release 2.0.0 GA? Since no code changes were required to make this migration, I suspect it might not matter much either way, but I think it's a good general question for us to answer.

Correct, I should have been more explicit. I was just thinking of the node version (14 vs 16), certainly not "update this example for every internal minor". If we went with the 16 option, if anything I'd update these for every internal major, and even that seems like maybe too much for non-GA versions.

Josmithr:
@ssimic2 Interested in your opinion on this. We should probably have an explicit policy on it.

Josmithr:
> > I think Alejandro's point was that our current GA release was developed on Node v14, rather than 16. So maybe we should not do this update until we release 2.0.0 GA? Since no code changes were required to make this migration, I suspect it might not matter much either way, but I think it's a good general question for us to answer.
> 
> Correct, I should have been more explicit. I was just thinking of the node version (14 vs 16), certainly not "update this example for every internal minor". If we went with the 16 option, if anything I'd update these for every internal major, and even that seems like maybe too much for non-GA versions.

Actually, now that Node 14 is no longer being supported, I think this is a change we should move forward on. But I think I agree that future migrations should be dependent on the highest common denominator between minimum Node support and what fluid-framework was using at the time of the release.

@alexvy86 @ssimic2 thoughts?

alexvy86:
> Actually, now that Node 14 is no longer being supported, I think this is a change we should move forward on. But I think I agree that future migrations should be dependent on the highest common denominator between minimum Node support and what fluid-framework was using at the time of the release.
> 
> @alexvy86 @ssimic2 thoughts?

Yeah, that makes sense.

PR #776, opened 2023-08-03
Increase timeout for react-starter-template tests

scottn12:
This PR increases the timeout for react-starter-template tests, since they were consistently timing out in automated test pipelines.

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -7,27 +7,27 @@ pool:
   vmImage: ubuntu-latest
 
 parameters:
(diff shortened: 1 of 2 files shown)


scottn12:
> LGTM.
> 
> I wonder if we should switch all our test to use local-driver instead of tinylicious for testing. Example: examples/view-integration/external-views/tests/index.ts

I think we probably should, it would also help speed up the build/test pipelines a lot

PR #779, opened 2023-09-11
Use node 18 on CI

CraigMacomber:
Update to a supported version of NodeJs for CI.


Josmithr (dismissed):

jason-ha (approved):
Should we bump `engines` in packages.json_s to say we are only supporting 18.x?

CraigMacomber:
> Should we bump `engines` in packages.json_s to say we are only supporting 18.x?

Yes: I'll drop support for node 16 and update the types and engines blocks in a follow-up.

PR #782, opened 2023-09-12
Update audience-demo

CraigMacomber:
Update deps and improve readme for  audience-demo.

Also updates CI from node 14 to node 16 to support `??=` syntax.

These updates might help with the node update (https://github.com/microsoft/FluidExamples/pull/779 ), but with the unrelated non-deterministic CI errors and no useful error output, that's just a guess.


CraigMacomber:
When running tests locally, I'm having issues with file watchers. Seems like tests shouldn't need file watchers, and non-test runs (ex: `npm start`) work fine. I can't see a way to disable file watchers for tests though, so I'll see if CI can run them.

robertobe-ms (approved):

Josmithr (approved):
🚀

PR #785, opened 2023-09-14
Misc Updates and cleanup

CraigMacomber:
Pull in updates from https://github.com/microsoft/FluidExamples/commit/61415de11293d8fddf1da342350f4e22a5d4845e except for updating the node version, then npm install in all the examples.

This mainly updates a lot of deps to help enable node 18, and fixes many minor issues (updates to address deprecation warning from the new versions, fix alphabetical ordering, mode dev deps to dev deps etc.)


angular-demo/package-lock.json
```diff
@@ -264,6 +264,28 @@
 				}
 			}
 		},
+		"node_modules/@angular-devkit/core/node_modules/ajv": {
+			"version": "8.9.0",
```
jason-ha, line 268: Seems odd to have a lock file only change including when lock file version isn't updated. But if all is working I can't say things are wrong.

node-demo/package.json
```diff
@@ -18,19 +19,22 @@
...
+		"node": ">=16.18.38"
+	},
 	"dependencies": {
-		"@fluidframework/tinylicious-client": "^1.3.0",
-		"fluid-framework": "^1.0.1",
+		"@fluidframework/tinylicious-client": "^1.3.7",
+		"fluid-framework": "^1.3.7",
 		"readline-sync": "^1.4.10",
 		"prettier": "^2.7.1"
 	},
```
jason-ha, line 30: Should only be a dev dependency. Okay not to correct.

jason-ha (approved):
One pre-existing dependency not corrected to devDependency as some others had. Okay to proceed with that or fix.

angular-demo/package-lock.json
CraigMacomber: It is oddL I just ran `npm i` in all the example folders, so maybe its related to a new npm version. I think I'll make a follow-up round of changes regenerating the lock files from scratch with latest version and dedup to clean up any odd state thats accumulated in here.

build.yml
```diff
@@ -68,30 +68,29 @@ steps:
...
-      - task: Npm@1
+          script: "npm run lint"
+          workingDirectory: ${{ parameter.Value }}
+      - task: CmdLine@2
         displayName: Test - ${{ parameter.Value }}
         inputs:
-          command: "custom"
-          workingDir: ${{ parameter.Value }}
-          customCommand: "run ci:test"
+          command: "npm run ci:test"
```
jason-ha, line 85: >command [](http://example.com/codeflow?start=10&length=7)

```suggestion
script
```

PR #788, opened 2023-09-15
Update build.yml

CraigMacomber:
Remaining changes from https://github.com/microsoft/FluidExamples/pull/785, cleaning up the build.yml. Includes a fix from that version which fixes test running.

This does not include the Node version update: that will follow in a separate change.


jason-ha (approved):
Looks correct to me

PR #790, opened 2023-09-15
Only support Node 18 and newer

CraigMacomber:
Require at least Node 18.

This removes support for older unsupported major versions of NodeJs.

### DIFF
--- a/collaborative-text-area/package-lock.json
+++ b/collaborative-text-area/package-lock.json
@@ -20,7 +20,7 @@
 				"@fluidframework/build-common": "^2.0.0",
 				"@testing-library/react": "^11.2.7",
 				"@testing-library/user-event": "^12.8.3",
-				"@types/node": "^16.18.38",
+				"@types/node": "^18.0.0",
 				"@types/react": "^17.0.34",
 				"@types/react-dom": "^17.0.11",
 				"jest": "^29.2.2",
@@ -33,7 +33,7 @@
 				"typescript": "^4.5.5"
 			},
 			"engines": {
-				"node": ">=16.18.38"
+				"node": ">=18.0.0"
 			}
 		},
 		"node_modules/@aashutoshrathi/word-wrap": {
@@ -3248,6 +3248,12 @@
 				"uuid": "^8.3.1"
 			}
 		},
+		"node_modules/@fluidframework/server-memory-orderer/node_modules/@types/node": {
+			"version": "16.18.51",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-16.18.51.tgz",
+			"integrity": "sha512-LKA7ZhY30I8PiUOzBzhtnIULQTACpiEpPXLiYMWyS+tPAORf+rmXUadHZXB/KFrFyMjeHeKc1GFvLd+3f7LE3w==",
+			"dev": true
+		},
 		"node_modules/@fluidframework/server-memory-orderer/node_modules/ws": {
 			"version": "7.5.9",
 			"resolved": "https://registry.npmjs.org/ws/-/ws-7.5.9.tgz",
@@ -3369,6 +3375,12 @@
 				"sillyname": "^0.1.0",
 				"uuid": "^8.3.1"
 			}
+		},
+		"node_modules/@fluidframework/server-services-core/node_modules/@types/node": {
+			"version": "16.18.51",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-16.18.51.tgz",
+			"integrity": "sha512-LKA7ZhY30I8PiUOzBzhtnIULQTACpiEpPXLiYMWyS+tPAORf+rmXUadHZXB/KFrFyMjeHeKc1GFvLd+3f7LE3w==",
+			"dev": true
 		},
 		"node_modules/@fluidframework/server-services-shared": {
 			"version": "1.0.1",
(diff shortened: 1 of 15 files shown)


collaborative-text-area/package-lock.json
```diff
@@ -3248,6 +3248,12 @@
 				"uuid": "^8.3.1"
 			}
 		},
+		"node_modules/@fluidframework/server-memory-orderer/node_modules/@types/node": {
+			"version": "16.18.51",
```
jason-ha, line 3252: Huh?
Why are any @types a non-dev dependency for any production package?
This is not a problem here. Hopefully there isn't any actual production use; because if there is then this is reaching back to 16.

Same for server-services-core a little below.

package.json
```diff
@@ -14,7 +14,7 @@
   },
   "homepage": "https://github.com/microsoft/FluidExamples#readme",
   "engines": {
-    "node": ">=16.18.38"
+    "node": ">=18.0.0"
```
jason-ha, line 17: >18.0.0 [](http://example.com/codeflow?start=15&length=6)

```suggestion
18.7.1
```
since that is what we are working with and validate?

jason-ha (approved):
:shipit:

collaborative-text-area/package-lock.json
CraigMacomber: There is some inconsistency around how to handle typed dependencies. If they are imported in non-test code, from a security perspective they are non-dev deps (and could contain not type code if not working as they should).

Also if those types are reexported from the public API of a package (which we don't have tools to detect), then typescript users of the package may need a matching version.

Now for examples/application (not libraries) who effectively have no public API at all, only the first of those points matters.

Anyway, for the specific case you mention, thats server code which uses node APIs (relies on polyfills for node APIs if run in the browser), and thus uses those types in its production code, so at least the first reason above applies. At least I think thats whats going on here.

CraigMacomber: I say there is inconsistency there since I don't think we (the Fluid project and/or TypeScript package devs more generally) have a clear policy and do the same thing everywhere: maybe we do though and I'm just not aware of it.

package.json
CraigMacomber: Where did you get that version? We now only validate node 20+, but at least the types dep should keep us from using features not in 18.0.0 directly.

jason-ha: Right - that is the version we had previously validated with.

jason-ha: It can be observed in pipeline logs.
[https://dev.azure.com/fluidframework/public/_build/results?buildId=190680&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=09732988-edb3-5e88-604f-3f1d71269f50](https://dev.azure.com/fluidframework/public/_build/results?buildId=190680&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=09732988-edb3-5e88-604f-3f1d71269f50)

CraigMacomber: If thats a requirement for the fluid server packages we consume, they can/should have their own engines block which will apply that restriction for any code that uses it which will be more robust, targeted and maintainable then making every use of it stay up to date via declaring its own limitations.

collaborative-text-area/package-lock.json
jason-ha: I looked for imports of those packages and found none. Maybe I missed them? There could be polyfills that do such imports, but then they should contain the imports. I am assuming polyfill packages being imported.
But it doesn't seem like `@types` are commonly usable like that. They are just metadata and never impact output javascript.

Definitely if your production lib uses a package then it needs to be a non-dev dependency.
devDependencies are appropriate for anything needed at package build time and those are still relevant to security concerns. But they don't need to be concerns for consumers of the package.

package.json
jason-ha: Agreed.
I was _suggesting_ the version here because that is what we had checked against. We expect but don't know that earlier versions of 18 are supported. We want people on 20+ so I don't think asking for recent 18 version is a problem.

jason-ha: BTW that log I referenced was FluidExamples - not FluidFramework proper.

CraigMacomber: I'd like to avoid rejecting users wanting to play with the examples because their node install is a bit old, so I just manually tested 18.0.0 locally and confirmed it works fine with at least the brainstorm sample. I also confirmed (before picking 18.0.0) that it had no patch releases, so it likely didn't have horrible problems.

collaborative-text-area/package-lock.json
CraigMacomber: I wouldn't be surprised if our server packages need some dependency pruning in this area. Some of the packages do have some node imports in production code (ex: https://github.com/microsoft/FluidFramework/blob/bdd587f594fb6f0d9c6b7068559386233660d37b/server/gitrest/packages/gitrest-base/src/utils/filesystems.ts#L6 ), but its possible the ones we see here don't.

When we have polyfills (or trivial stubs for node APIs), usually we are pulling them in via webpack config: My point was not that the @types do it (they don't), but rather we have cases where using @types/node make sense in prod code run in the browser since that @types/node code is providing types which are implemented by something other than node.

jason-ha: `@types/node` is not at all `node`. `@types/*` packages only contain type metadata (or at least that is all they are supposed to contain). I have never seen a production use of a `@types/*` package. I'll submit a FF PR and discussion can happen there. I expect these are all oversights.

PR #791, opened 2023-09-18
Updating root README

WillieHabi:
`npm run start` does not start local server for any of the examples. Must run npm start:server first to start local azure service/tinylicious

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To get started, clone the repository and navigate to one of the example folders.
 From the example root, you can install and start the example.
 
 1. Run `npm install` from the example root
-2. Run `npm start server` to start the local server
+2. Run `npm start:server` to start the local server
 3. Run `npm start` to start the client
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,13 +12,14 @@ To get started, clone the repository and navigate to one of the example folders.
 From the example root, you can install and start the example.
 
 1. Run `npm install` from the example root
-2. Run `npm run start` to start both the client and server
+2. Run `npm start:server` to start the local server
+3. Run `npm start` to start the client
```
scottn12, line 16: Which example does this actually run? We should probably specify that in the README too

WillieHabi: I think these instructions work with every example in the repo, if I'm not mistaken

scottn12: Ah gotcha! That's on me for misunderstanding 😅

ChumpChief:
Node-demo has since been removed from the repo, so closing this PR.

PR #797, opened 2023-10-30
Standardize examples to use @fluidframework/azure-local-service instead of tinylicious

scottn12:
Standardizes all examples to use `@fluidframework/azure-local-service` instead of tinylicious

### DIFF
--- a/angular-demo/package.json
+++ b/angular-demo/package.json
@@ -11,7 +11,7 @@
 		"ci:test:jest": "jest --ci --reporters=default --reporters=jest-junit",
 		"ng": "ng",
 		"start": "ng serve",
-		"start:server": "npx tinylicious",
+		"start:server": "npx @fluidframework/azure-local-service",
 		"build": "ng build",
 		"watch": "ng build --watch --configuration development",
 		"test": "start-server-and-test start:server 7070 test:jest",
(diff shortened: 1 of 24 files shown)


ChumpChief:
We should probably normalize on `@fluidframework/azure-local-service` since that's what we advertise to customers to use.

scottn12:
> We should probably normalize on `@fluidframework/azure-local-service` since that's what we advertise to customers to use.

@ChumpChief, do you know where we advertise to customers to use azure-local-service? I can't really find it on ff.com and it looks like we encourage using either tinylicious or just FRS [here](https://fluidframework.com/docs/deployment/service-options/).

ChumpChief:
> @ChumpChief, do you know where we advertise to customers to use azure-local-service? I can't really find it on ff.com and it looks like we encourage using either tinylicious or just FRS [here](https://fluidframework.com/docs/deployment/service-options/).

https://learn.microsoft.com/en-us/azure/azure-fluid-relay/how-tos/local-mode-with-azure-client#running-azure-fluid-relay-service-locally

angular-demo/package.json
```diff
@@ -40,6 +40,7 @@
 		"@angular-devkit/build-angular": "~13.3.9",
 		"@angular/cli": "~13.3.9",
 		"@angular/compiler-cli": "~13.3.11",
+		"@fluidframework/azure-local-service": "^1.1.0",
```
ChumpChief, line 43: If we're doing npx, then shouldn't need to include this as a dependency, same for other instances in other demos.

ChumpChief (approved):

PR #802, opened 2023-11-28
converts pipeline to M365 template

seanimam:
This is not to be merged in its current state, just for testing the pipeline conversion update

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -3,94 +3,111 @@
 
 # FluidExamples build for CI and PR
 
-pool:
-  vmImage: ubuntu-latest
-
 parameters:
-  - name: angular-demo
-    type: string
-    default: angular-demo
-  - name: audience-demo
-    type: string
-    default: audience-demo
-  - name: brainstorm
-    type: string
-    default: brainstorm
-  - name: collaborative-text-area
-    type: string
-    default: collaborative-text-area
-  - name: multi-framework-diceroller
-    type: string
-    default: multi-framework-diceroller
-  - name: node-demo
-    type: string
-    default: node-demo
-  - name: react-demo
-    type: string
-    default: react-demo
-  - name: react-starter-template
-    type: string
-    default: react-starter-template
-
+- name: angular-demo
+  type: string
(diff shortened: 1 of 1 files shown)


build.yml
```diff
@@ -31,66 +28,89 @@ parameters:
...
+  repositories:
+    - repository: m365Pipelines
+      type: git
+      name: 1ESPipelineTemplates/M365GPT
+      ref: refs/tags/release
+extends:
+  template: v1/M365.Official.PipelineTemplate.yml@m365Pipelines
+  parameters:
+    pool:
+      name: NewLarge-linux-1ES
```
alexvy86, line 54: For building the examples repo I think it might be preferable to use the Small pool. Unless the build for this repo does particularly compute-intensive things, but I doubt it.

alexvy86 (commented):
The changes seem fine overall... but do we even need to convert this pipeline in particular? It produces no artifacts. I thought the move to the 1ES templates was mostly so the output artifacts complied with any necessary checks and analysis. cc @rohandubal 

One nit: I'd restore some of the blank lines that the tool removed, especially the ones close to the top of the file separating the "main root sections".

build.yml
seanimam: What is the name of the small pool?

alexvy86: I think `Small-linux-1ES`, Rohan can confirm.

build.yml
```diff
@@ -49,48 +46,77 @@ pr:
...
+      sourceAnalysisPool:
+        name: Azure-Pipelines-1ESPT-ExDShared
+        image: windows-2022
+        os: windows
+    customBuildTags:
+      - ES365AIMigrationTooling
+    stages:
+      - stage: stage
+        jobs:
+          - job: job
```
alexvy86, line 79: Nit
```suggestion
      - stage: run_build
        jobs:
          - job: run_build
```

build.yml
seanimam: Its Small-1ES, I have updated the PR to this.

PR #808, opened 2024-01-24
Version update

nmsimons:
Bump version and switch to using tinylicious


brainstorm/package.json
```diff
@@ -15,7 +15,7 @@
 		"lint": "eslint src",
 		"webpack": "cross-env FLUID_CLIENT='azure' webpack",
 		"start": "npm run dev",
-		"start:server": "npx @fluidframework/azure-local-service"
+		"start:server": "npx tinylicious@latest"
```
ChumpChief, line 18:
```suggestion
		"start:server": "npx @fluidframework/azure-local-service@2.0.0-rc.1.0.1"
```

item-counter/package.json
```diff
@@ -15,7 +15,7 @@
 		"lint": "eslint src",
 		"webpack": "cross-env FLUID_CLIENT='azure' webpack",
 		"start": "npm run dev",
-		"start:server": "npx @fluidframework/azure-local-service"
+		"start:server": "npx tinylicious@latest"
```
ChumpChief, line 18:
```suggestion
		"start:server": "npx @fluidframework/azure-local-service@2.0.0-rc.1.0.1"
```

ChumpChief (commented):
Let's do this instead which should be a more correct solution.

Problem seems to be that the "latest" tagged `@fluidframework/azure-local-service` is 1.1.1 (whereas "latest" tagged tinylicious is 3.0.0, which is why it also fixes).  Not entirely sure what changed since we released that would have caused the problem to arise (did we change the tagging or something?).

ChumpChief (approved):

PR #878, opened 2024-04-05
Fix drag and drop issues with groups

nmsimons:


brainstorm/src/react/noteux.tsx
```diff
@@ -242,13 +240,18 @@ function NoteToolbar(props: { note: Note; clientId: string; notes: Items }): JSX
...
 export function AddNoteButton(props: { parent: Items; clientId: string }): JSX.Element {
 	const [{ isActive }, drop] = useDrop(() => ({
-		accept: dragType.NOTE,
+		accept: [dragType.NOTE, dragType.GROUP],
 		collect: (monitor) => ({
 			isActive: monitor.canDrop() && monitor.isOver(),
 		}),
+		canDrop: (item) => {
+			if (item instanceof Note) return true;
+			if (item instanceof Group && item.items !== props.parent) return true;
```
CraigMacomber, line 249: I think you need a loop here to check parents all the way up the tree, since you shouldn't be able to put something inside a group inside itself.

Seems like a Tree.isUnder(parent, child) or something like that would be useful here. Maybe a good feature request?

nmsimons: I have added a better test.

brainstorm/src/utils/utils.ts
```diff
@@ -50,3 +50,21 @@ export enum selectAction {
 	REMOVE,
 	SINGLE,
 }
+
+// function that tests to see if a target
+// is anywhere within a parent's subtree
```
CraigMacomber, line 55: You should use a doc comment so developers mousing over usages of this can read the comment in IntelliSense.

Additionally saying the function is a function isn't adding any value, and can be left put.

This is documented in terms of "parent" but the signature says "Group".

This is a very inefficient way to implement this. Instead of searching everything under the group recursively, you should walk the parents of the target checking for group. THen its O(depth) not O(size of subtree) and also works for all kinds of nodes, not just Groups and Items.

I'm not sure the name isCircular is good, since this does not check for a cycle to exist, but instead is checking if target is under group.

You could name it something like "isUnder" or "moveWouldBeCircular"

CraigMacomber (approved):

CraigMacomber:
An API to help with this is proposed in https://github.com/microsoft/FluidFramework/pull/20537

PR #945, opened 2024-05-16
Update to rc4

nmsimons:
Version bump that required some minor code changes.


brainstorm/src/infra/azure/azureTokenProvider.ts
```diff
@@ -123,8 +123,8 @@ export class InsecureTokenProvider implements ITokenProvider {
 export const user = generateUser();
 
 export const azureUser = {
-	userId: user.id,
-	userName: user.name,
+	id: user.id,
+	name: user.name,
 };
```
ChumpChief, line 128: Maybe consolidate user/azureUser together now that they look the same?  Same for item-counter.

ChumpChief (approved):
One optional suggestion, looks good tho!

brainstorm/src/infra/azure/azureTokenProvider.ts
nmsimons: I'll do that next time I actually touch code. I want to get this in today.

PR #974, opened 2024-06-20
Update to rc5

nmsimons:
Will not merge until 5.0.1 since I found a bug while doing this. But the code is correct as far as I can tell.


brainstorm/src/app_load.tsx
```diff
@@ -1,11 +1,11 @@
-import { OdspClient } from "@fluid-experimental/odsp-client";
+import { OdspClient } from "@fluidframework/odsp-client";
```
ChumpChief, line 1: Nit: re-sort package name alphabetically here and elsewhere now that the scope changed

brainstorm/src/app_load.tsx
```diff
@@ -20,8 +20,14 @@ export async function loadApp(
 	const { services, container } = await loadFluidData(containerId, containerSchema, client);
 
 	// Initialize the SharedTree DDSes
-	const sessionTree = container.initialObjects.sessionData.schematize(sessionTreeConfiguration);
-	const appTree = container.initialObjects.appData.schematize(appTreeConfiguration);
+	const sessionTree = container.initialObjects.sessionData.viewWith(sessionTreeConfiguration);
+	if (sessionTree.compatibility.canInitialize) {
```
ChumpChief, line 24: What happens if it can't initialize, or is that a realistic risk?

ChumpChief (approved):
Looks good to me!

brainstorm/src/react/noteux.tsx
```diff
@@ -78,6 +78,16 @@ export function NoteView(props: {
 		return unsubscribe;
 	}, []);
 
+	// Register for tree deltas when the component mounts.
+	// Any time the node changes, the app will update
+	useEffect(() => {
+		// Returns the cleanup function to be invoked when the component unmounts.
+		const unsubscribe = Tree.on(props.note, "nodeChanged", () => {
+			setInvalidations(invalidations + Math.random());
```
ChumpChief, line 86: This setInvalidations pattern is pretty unpleasant - I'm not going to block on it since view integration isn't exactly the point of these examples and I'd like to see RC5 integration in as a higher priority, but would be good to establish a more-recommendable usage pattern for SharedTree.

ChumpChief (approved):

brainstorm/src/app_load.tsx
nmsimons: fixed

brainstorm/src/react/noteux.tsx
nmsimons: Fixing this will require some significant changes. Good next week project.

rohandubal (approved):

PR #1032, opened 2024-08-13
test: Replace `jest` + `puppeteer` for testing with `playwright` and add interactive tests

Josmithr:
[Playwright](https://playwright.dev/) is a more modern alternative to `jest` and friends. It provides a more user-friendly test-writing ecosystem, and makes debugging tests **much** easier.

Note that this PR substantially reduces the amount of test boilerplate we have, while expanding functionality.

Other benefits:
- Supports ESM out of the box (no need to opt into experimental jest features that have been "experimental" for many years with no movement.
- Has great docs (see link above)
- Microsoft-maintained (and _actively_ maintained, unlike `jest` and friends).
- Allows (easily) launching multiple dev servers prior to running tests, allowing us to remove our `start-server-and-test` pattern (and dev dependency).
- Supports screenshot-based visual regression testing for UI validation.
    - For later - would require setting up Docker infra to ensure consistent visual output across environments.

Other notes:
- `item-counter-spe` does not currently have runnable tests. `odsp-client` doesn't have a "local" mode, so this app was never set up with a usable test flow. It contained a test module (copied from `item-counter`), but those tests were never runnable. This PR removes the unused test dependencies and infra in that package, but does not set up `playwright`. That can be added in the future when `odsp-client` supports a "local" mode, and the app is configured to leverage it.


brainstorm/test/App.test.ts
```diff
@@ -0,0 +1,68 @@
...
+		await page.goto("/", { waitUntil: "domcontentloaded" });
+		expect(await page.title()).toBe("Brainstorm Demo");
+		const collaborationUrl = await page.url();
+
+		// Navigate to the collaboration URL - load existing Container case
+		await page.goto(collaborationUrl, { waitUntil: "domcontentloaded" });
+	});
+
+	test("Add group", async ({ page }) => {
+		let groups = await page.getByLabel("Note Group");
```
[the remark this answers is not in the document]
Josmithr: I reassign these later in the test (checking the state before and after interactions). These _could_ be separate variables, but they really represent the same content, just at different times.

brainstorm/package.json
```diff
@@ -7,19 +7,18 @@
...
-		"ci:test": "start-server-and-test start:server 7070 ci:test:jest",
-		"ci:test:jest": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --ci --reporters=default --reporters=jest-junit",
+		"ci:test": "npm run test",
 		"compile": "tsc -b",
 		"dev": "cross-env webpack-dev-server",
 		"dev:local": "cross-env FLUID_CLIENT='local' webpack-dev-server",
 		"dev:azure": "cross-env FLUID_CLIENT='azure' webpack-dev-server",
 		"dev:spe": "cross-env FLUID_CLIENT='spe' webpack-dev-server",
 		"format": "prettier src --write",
+		"postinstall": "npx playwright install --with-deps",
```
Josmithr, line 17: Ensure `playwright` browsers are installed.

brainstorm/playwright.config.ts
```diff
@@ -0,0 +1,65 @@
...
+	fullyParallel: true,
+
+	// Fail the build on CI if you accidentally left test.only in the source code.
+	forbidOnly: !!process.env.CI,
+
+	// Retry on CI only.
+	retries: process.env.CI ? 2 : 0,
+
+	// Opt out of parallel tests on CI.
+	workers: process.env.CI ? 1 : undefined,
```
Josmithr, line 23: This probably isn't necessary - came with the "default" config in `playwright`'s docs.

brainstorm/playwright.config.ts
```diff
@@ -0,0 +1,65 @@
+import { defineConfig, devices } from "@playwright/test";
+
+const frontendPort = 8080;
+const frontEndUrl = `http://localhost:${frontendPort}`;
+
+const backendPort = 7070;
+const backEndUrl = `http://localhost:${backendPort}`;
+
+export default defineConfig({
+	// Look for test files in the "tests" directory, relative to this configuration file.
```
Josmithr, line 10:
```suggestion
	// Look for test files in the "test" directory, relative to this configuration file.
```

Josmithr:
Test results continue to be published as before:
image

brainstorm/test/App.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	});
+
+	test("Add note", async ({ page }) => {
+		let notes = await page.getByLabel("Note");
+		expect(notes).toHaveCount(0);
+
+		// Click the "Add Group" button
+		await page.getByText("Add Note").click();
+
+		// Verify that a group was added.
```
alexvy86, line 47:
```suggestion
		// Click the "Add Note" button
		await page.getByText("Add Note").click();

		// Verify that a note was added.
```

brainstorm/test/App.test.ts
```diff
@@ -0,0 +1,68 @@
...
+		notes = await page.getByLabel("Note");
+		expect(notes).toHaveCount(1);
+	});
+
+	test("Delete note", async ({ page }) => {
+		// Click the "Add Group" button
```
alexvy86, line 53:
```suggestion
		// Click the "Add Note" button
```

brainstorm/test/App.test.ts
```diff
@@ -0,0 +1,68 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
```
alexvy86, line 9:
```suggestion
/*!
 * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
 * Licensed under the MIT License.
 */
```

item-counter/test/App.test.ts
```diff
@@ -0,0 +1,47 @@
...
+	let collaborationUrl: string | undefined;
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto("http://localhost:8080", { waitUntil: "domcontentloaded" });
+		expect(await page.title()).toBe("Fluid Demo");
+		collaborationUrl = await page.url();
+	});
+
+	test("Load the container (smoke test)", async ({ page }) => {
+		assert(collaborationUrl !== undefined);
```
alexvy86, line 23: Nit: stick to `expect()` instead of using node's assert too?

item-counter/test/App.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+		// Click the "Insert" button 3 times
+		await page.click("text=Insert", { clickCount: 3 });
+
+		// Verify the counter value
+		let itemCountElement = await page.getByLabel("Item count");
+		let itemCount = Number.parseInt(await itemCountElement.innerText());
+		expect(itemCount).toEqual(3);
+
+		// Click the "Remove" button 3 times
```
alexvy86, line 39: I also wouldn't mind removing many of the comments, the test code feels self-explanatory enough.
```suggestion
		// Click the "Remove" button 2 times
```

brainstorm/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+	testDir: "test",
+
+	// Run all tests in parallel.
+	fullyParallel: true,
+
+	// Fail the build on CI if you accidentally left test.only in the source code.
+	forbidOnly: !!process.env.CI,
+
+	// Retry on CI only.
+	retries: process.env.CI ? 2 : 0,
```
alexvy86, line 20: This is tricky, Tony has argued a few times against retries like this in tests because they could hide flakiness. I think I'd start without retries.

item-counter/test/App.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+test.describe("item-counter", () => {
+	/**
+	 * Tracks the collaboration URL that is generated when navigating to the default URL.
+	 * Includes the generated container ID.
+	 */
+	let collaborationUrl: string | undefined;
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto("http://localhost:8080", { waitUntil: "domcontentloaded" });
```
alexvy86, line 17: This should work thanks to the config, right?
```suggestion
		await page.goto("/", { waitUntil: "domcontentloaded" });
```

alexvy86 (commented):
I like it! Some comments below.

brainstorm/playwright.config.ts
Josmithr: Yeah, I was leaning the same way. This was in the "recommended" base config from the website, so I figured I would leave it be initially and see what folks think. But I tend to agree.

item-counter/test/App.test.ts
Josmithr: Unfortunately, `expect` doesn't do type-narrowing, so using it would force us to use non-null operators everywhere 😕

item-counter/test/App.test.ts
Josmithr: Since these are public-facing examples, I erred on the side of excess documentation. If we feel strongly that these should be removed, I can, but the step-by-step explanation seems potentially useful.

alexvy86 (approved):

seanimam (dismissed):

brainstorm/package.json
tylerbutler: Is this the recommended approach? Does it do some smart caching so it's fast? How does it impact install times?

tylerbutler: Would every package that uses playwright run these? should they be at the root?

brainstorm/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
+import { defineConfig, devices } from "@playwright/test";
+
+const frontendPort = 8080;
+const frontEndUrl = `http://localhost:${frontendPort}`;
+
+const backendPort = 7070;
+const backEndUrl = `http://localhost:${backendPort}`;
```
tylerbutler, line 7: I think these need to be configurable and read the value set by assign-tsest-ports.

item-counter/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+		trace: "on-first-retry",
+
+		// Generate screenshots when a test fails
+		screenshot: "only-on-failure",
+	},
+	// Configure projects for major browsers.
+	projects: [
+		{
+			name: "chromium",
+			use: { ...devices["Desktop Chrome"] },
```
tylerbutler, line 44: Do we want to test all the browsers we officially support? Maybe later in a seaprate pipeline?

Josmithr: Possibly. I was leaving things simple for now, but we can expand if we'd like!

brainstorm/playwright.config.ts
Josmithr: I think that's true in `fluid-framework`, but these values were already hard-coded in this repo.

brainstorm/package.json
Josmithr: Depends on the setup. In this repo, since the packages are managed independently, doing it here makes sense. It does cache, so subsequent installs are fast (if no updates). We may want to do something different in `fluid-framework`, but I think this approach makes sense here.

The docs generally recommend devs run this as needed, and have CI run it before running tests. But since the results are cached, this seemed like an unnecessary concept hurdle to add here. Again, for `fluid-framework`, we may wish to do things differently. You and I should chat 🙂

Josmithr: A possible alternative would be to bind this to `pretest` instead of `postinstall`. Then the cost wouldn't be incurred until test time. Interested in your thoughts.

RishhiB (dismissed):

brainstorm/package.json
Josmithr: Actually, the more I think about it, the more I think that's a better way to do this. I'll make that update. Let me know if you (or anyone) disagrees :)

brainstorm/playwright.config.ts
tylerbutler: Ahh, got it. May be worth consolidating patterns we use, but such work would be separate anyway.

brainstorm/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	// Don't retry on failure.
+	retries: 0,
+
+	// Reporter to use
+	reporter: [
+		// Console output
+		["line"],
+		// JUnit XML report file output for CI
+		["junit", { outputFile: "test-results/junit-report.xml" }],
```
tylerbutler, line 27: As noted in my vitest PR, the pipelines in FluidFramework may need updates to use paths like this. Not relevant here necessarily but something to note.

brainstorm/package.json
tylerbutler: Honestly I kinda forgot this was FluidExamples :) I like pretest instead - should also work with fluid-build in the main repo.

tylerbutler (approved):
Love the direction! Thanks for pushing this forward.

PR #1127, opened 2024-09-27
Make sample SPE app work for single-tenant Entra Apps

alexvy86:
I believe this change still allows the app to live in the "common" Entra tenant owned by Microsoft, but now would require the user to pass its id explicitly.

I had to make this change to make the app work with an M365/Entra tenant from my developer account.

### DIFF
--- a/item-counter-spe/README.md
+++ b/item-counter-spe/README.md
@@ -8,16 +8,24 @@ This app is designed to use SharePoint Embedded (SPE). The core application code
 this sample uses odsp-client to connect to Fluid Framework as opposed to azure-client. The differences are primarilly isolated to the
 infra folder, although the initialization flow for this app is different in that it includes auth while the AFR samples are anonymous.
 
-To use SPE you need to create an M365 developer account and configure SharePoint Embedded. The easiest way to get started is to install the SharePoint Embedded Visual Studio Code extension. From there you can create a new M365 developer tenant, create a new client app (with the require Microsoft Entra client ID) create new File Storage Container Types, and create new File Storage Containers.
+To use SPE you need to create an M365 developer account and configure SharePoint Embedded.
+The easiest way to get started is to install the SharePoint Embedded Visual Studio Code extension.
+From there you can create a new M365 developer tenant (which includes creating an Entra tenant), create a new client app
+(with the required Microsoft Entra client ID), create new File Storage Container Types, and create new File Storage Containers.
 
-This sample app requires that you have a Microsoft Entra client Id, have created a File Storage Container Type, and that the tenant you plan to use has a File Storage Container with that File Storage Container Type ID. Once you have done that, create a .env file in the item-counter-spe folder with the following content:
+This sample app requires that you have a Microsoft Entra tenant, a Microsoft Entra client Id for the app, have created
+a File Storage Container Type, and that the tenant you plan to use has a File Storage Container with that File Storage
+Container Type ID.
+Once you have done that, create a `.env` file in the item-counter-spe folder with the following content:
 
-```
+```plaintext
 SPE_CLIENT_ID='your client id'
 SPE_CONTAINER_TYPE_ID='your container type id'
+SPE_ENTRA_TENANT_ID='your entra tenant id
 ```
 
-With that in place, you can run this sample (`npm run dev`). Log in with the admin credentials for the tenant.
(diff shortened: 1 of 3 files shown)


item-counter-spe/src/index.tsx
```diff
@@ -23,8 +23,7 @@ async function start() {
 	// If the tokenResponse is not null, then the user is signed in
 	// and the tokenResponse is the result of the redirect.
 	if (tokenResponse !== null) {
-		const account = msalInstance.getAllAccounts()[0];
-		signedInStart(msalInstance, account);
+		await signedInStart(msalInstance, tokenResponse.account);
```
alexvy86, line 26: The missing await here and below seems like an oversight. But more importantly, if we have a token response then we already have the account, no need to default to the first one. Which in fact might even be a bug? If I'm logged in to the account in the Entra tenant for the app but also my account on the common Msft tenant, could the first one in the list be the incorrect one for the app...?

item-counter-spe/README.md
```diff
@@ -8,16 +8,24 @@ This app is designed to use SharePoint Embedded (SPE). The core application code
...
+This sample app requires that you have a Microsoft Entra tenant, a Microsoft Entra client Id for the app, have created
+a File Storage Container Type, and that the tenant you plan to use has a File Storage Container with that File Storage
+Container Type ID.
+Once you have done that, create a `.env` file in the item-counter-spe folder with the following content:
 
-```
+```plaintext
 SPE_CLIENT_ID='your client id'
 SPE_CONTAINER_TYPE_ID='your container type id'
+SPE_ENTRA_TENANT_ID='your entra tenant id
```
alexvy86, line 24:
```suggestion
SPE_ENTRA_TENANT_ID='your entra tenant id'
```

item-counter-spe/src/infra/authHelper.ts
```diff
@@ -9,12 +9,17 @@ export async function authHelper(): Promise<PublicClientApplication> {
...
+	}
+
 	// Create the MSAL instance
 	const msalConfig = {
 		auth: {
 			clientId,
-			authority: "https://login.microsoftonline.com/common/",
-			tenantId: "common",
+			authority: `https://login.microsoftonline.com/${tenantId}/`,
+			tenantId
```
kashms, line 22: Ah yes, this was the other change I had to make

nmsimons (approved):

item-counter-spe/README.md
```diff
@@ -8,16 +8,24 @@ This app is designed to use SharePoint Embedded (SPE). The core application code
 this sample uses odsp-client to connect to Fluid Framework as opposed to azure-client. The differences are primarilly isolated to the
 infra folder, although the initialization flow for this app is different in that it includes auth while the AFR samples are anonymous.
 
-To use SPE you need to create an M365 developer account and configure SharePoint Embedded. The easiest way to get started is to install the SharePoint Embedded Visual Studio Code extension. From there you can create a new M365 developer tenant, create a new client app (with the require Microsoft Entra client ID) create new File Storage Container Types, and create new File Storage Containers.
+To use SPE you need to create an M365 developer account and configure SharePoint Embedded.
+The easiest way to get started is to install the SharePoint Embedded Visual Studio Code extension.
+From there you can create a new M365 developer tenant (which includes creating an Entra tenant), create a new client app
+(with the required Microsoft Entra client ID), create new File Storage Container Types, and create new File Storage Containers.
```
krauders, line 14: Were you actually able to create File Storage Container Types and File Storage Containers using the VSCode extension? It didn't work for me, I could only make a "trial" container type, and I had to use the azure function sample app to do it.

alexvy86: Yes, but the container I created from the VSCode extension didn't show up when listing containers in the azure function sample app 🤔. The one I created in the app showed up in VSCode without issues, but the app never shows the one from VSCode. I'm not sure which of the two is being used when I run the item-counter-spe example, I'll try deleting both and confirm this app fails, then create a container from VSCode and see if that's enough for it to work.

kashms: You can also see the container type ID in the VScode extension and see what you're using in the sample app's env file

krauders: I had the same issue. The containers created in VSCode didn't show up when the app queries for containers. That's why I had to use the Azure app sample.

PR #1159, opened 2024-10-08
build: Update dependencies

Josmithr:


ChumpChief:
I've been trying to get dependabot to be more helpful here - was it not detecting some of these dependencies?

ChumpChief (approved):

Josmithr:
> I've been trying to get dependabot to be more helpful here - was it not detecting some of these dependencies?

Not all of them. But I also like to run local validation when bumping dependencies anyways, so I usually just do it myself. I did link the PRs that I saw that were updating a subset of these deps though. I'll close those after merging this.

ChumpChief:
It may have also been hitting PR limits - we might consider adding more groupings (maybe the playwright dependencies should be grouped?) to avoid maxing out, or we could just raise the limit.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

PR #1161, opened 2024-10-10
Avoid needless npx, do an npm audit fix, improve readme

CraigMacomber:
Use dev dependencies instead of npx so versions are controlled by package lock.

Ran npm audit to resolve three security vulnerabilities.

Add note to readme about dev server output.

### DIFF
--- a/item-counter/README.md
+++ b/item-counter/README.md
@@ -19,6 +19,8 @@ Running this command from your terminal window will launch the Azure Fluid Relay
 npm run start
 ```
 
+This command starts the webpack development server, which will make the application available at [http://localhost:8080/](http://localhost:8080/).
+
 One important note is that you will need to use a token provider or, purely for testing and development, use the insecure token provider. There are instructions on how to set this up on the [Fluid Framework website](https://aka.ms/fluid).
 
 All the code required to set up the Fluid Framework and SharedTree data structure is in the infra folder. Most of this code will be the same for any app.
--- a/item-counter/package.json
+++ b/item-counter/package.json
@@ -15,8 +15,8 @@
 		"format": "prettier src --write",
 		"lint": "eslint src",
 		"start": "npm run dev",
-		"start:server": "npx @fluidframework/azure-local-service@^2.0.1",
-		"pretest": "npx playwright install --with-deps",
+		"start:server": "azure-local-service",
+		"pretest": "playwright install --with-deps",
 		"test": "playwright test",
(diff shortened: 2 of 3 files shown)


item-counter/package.json
```diff
@@ -40,6 +40,7 @@
 	"devDependencies": {
 		"@eslint/eslintrc": "^3.1.0",
 		"@eslint/js": "^9.12.0",
+		"@fluidframework/azure-local-service": "^2.3.1",
```
Josmithr, line 43: @ChumpChief I am personally a fan of this pattern over npx invocations. But I seem to remember this being a deliberate choice in the past. Where do we currently stand on this?

CraigMacomber: brainstorm already has this as a dev dependency (though also needlessly uses npx).

Generally, I like having it as a dev dep so we don't prompt to user to get permission to install the service, then also not have control over what version gets used.

item-counter/README.md
```diff
@@ -19,6 +19,8 @@ Running this command from your terminal window will launch the Azure Fluid Relay
 npm run start
 ```
 
+This command starts the webpack development server, which will make the application available at [http://localhost:8080/](http://localhost:8080/).
```
Josmithr, line 22: Can we add this note to the `item-counter-spe` README as well?

item-counter/package.json
```diff
@@ -15,8 +15,8 @@
 		"format": "prettier src --write",
 		"lint": "eslint src",
 		"start": "npm run dev",
-		"start:server": "npx @fluidframework/azure-local-service@^2.0.1",
-		"pretest": "npx playwright install --with-deps",
+		"start:server": "azure-local-service",
+		"pretest": "playwright install --with-deps",
```
Josmithr, line 19: Can we make this change to `pretest` in `brainstorm` as well?

item-counter/README.md
CraigMacomber: We can add it to brainstorm. but item-counter-spe never mentions `npm run start` so I don't know how this applies to it. I don't actually know what running that app is like since I haven't bothered setting up a developer account to be able to run it.

item-counter/package.json
ChumpChief: Seems fine - consistency matters more to me here than which is used.

item-counter/package.json
CraigMacomber: done in https://github.com/microsoft/FluidExamples/pull/1162

Josmithr (approved):

PR #1471, opened 2025-05-19
Fix usage of tailwindcss v4, update instructions, regenerate css

steffenloesch:
The css files and commands in the README.md files were not working for tailwindcss v4, they were still tailored for v3. This PR fixes that and regenerates the css files with v4.

See also <https://tailwindcss.com/docs/installation/tailwind-cli>.

This also fixes a small typo in the item-counter-spe Readme file, and gives more accurate paths to files in all Readmes.


brainstorm/src/output.css
```diff
@@ -1,1305 +1,767 @@
...
-	transition-duration: 150ms;
-}
-
-.duration-500 {
-	transition-duration: 500ms;
-}
-
-.ease-out {
-	transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
+/*! tailwindcss v4.1.7 | MIT License | https://tailwindcss.com */
```
Josmithr, line 1: We should probably make sure these tailwind-generated files are ignored by prettier. Would you mind adding `.prettierignore` files that ensure these are ignored by the formatter?

Josmithr (dismissed):
Left a couple of small-ish suggestions but otherwise looks great! Thanks for cleaning this up!

brainstorm/src/output.css
steffenloesch: Good call! `npm run format` was heavily formatting these. I've added the `.prettierignore` files now.

nmsimons (approved):

Josmithr (approved):

PR #1635, opened 2025-09-24
Canvas

nmsimons:
Add a new example that includes extensive use of presence and SharedTree.


canvas/src/presence/Interfaces/PresenceManager.ts
```diff
@@ -0,0 +1,68 @@
...
+ * - Real-time state synchronization
+ * - Multi-client coordination
+ * - Event-driven updates
+ */
+
+import {
+	LatestRaw,
+	LatestRawEvents,
+	LatestMapRaw,
+	LatestMapRawEvents,
```
jason-ha, line 16: We should probably use the validated version of API in our sample.
Probably TypeBox is good solution. [Presence readme](https://github.com/microsoft/FluidFramework/blob/main/packages/framework/presence/README.md) has [example](https://github.com/microsoft/FluidFramework/blob/main/packages/framework/presence/README.md#example-validated-setup).

If not, then we should comment why this example isn't even though validated pattern is our recommendation.

jason-ha (commented):
Quick notes from brief look

canvas/src/index.css
```diff
@@ -0,0 +1,198 @@
+@import "tailwindcss";
+
+/*
```
tylerbutler, line 3: Do you need the commented out styles?

canvas/src/react/components/app/App.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+	useEffect(() => {
+		// View changed
+	}, [view]);
+
+	// Initialize iOS Safari z-index fixes
+	useEffect(() => {
+		// Temporarily disabled to debug toolbar visibility
+		// fixIOSZIndexIssues();
+	}, []);
```
tylerbutler, line 171: Still needed?

canvas/src/start/azureStart.ts
```diff
@@ -0,0 +1,173 @@
...
+		const urlParams = new URLSearchParams(window.location.search);
+		let containerId = urlParams.get("id") ?? "";
+
+		// Initialize Devtools logger if in development mode
+		const logger = undefined;
+		// Commented out to suppress FluidDevToolsLogger console messages
+		// if (process.env.NODE_ENV === "development") {
+		// 	const { createDevtoolsLogger } = await import("@fluidframework/devtools/beta");
+		// 	logger = createDevtoolsLogger();
+		// }
```
tylerbutler, line 147: Still needed?

canvas/EVENT_SUBSCRIPTION_CONSOLIDATION.md
tylerbutler, line 1: This looks like a dev doc - maybe move this and similar ones into a docs folder or delete? I think they'll be confusing to people reading the example.

canvas/src/infra/azure/azureTokenProvider.ts
```diff
@@ -0,0 +1,240 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { AzureMember, ITokenProvider, ITokenResponse, IUser } from "@fluidframework/azure-client";
+import { ScopeType } from "@fluidframework/protocol-definitions";
+import axios from "axios";
```
tylerbutler, line 8: It still bugs me that we use axios instead of fetch, but probably not in scope for now.

PR #1857, opened 2026-04-13
fix(item-counter): lint and type fixes

tylerbutler:
## Summary

- Update tsconfig jsx from `"react"` to `"react-jsx"` (React 19 automatic transform)
- Replace non-null assertions (`!`) with `?? ""` fallbacks
- Convert ternary expression to nullish coalescing (`??`)
- Fix import ordering per import-x/order
- Apply consistent-type-imports (separate type imports)
- Split mixed type/value imports for correct TypeScript compilation

These are independently valid code improvements, split out from the eslint-config-fluid adoption PR.

## Test plan

- [ ] `npm run compile` in item-counter has no new TypeScript errors
- [ ] Dev server starts correctly

### DIFF
--- a/item-counter/src/infra/clientProps.ts
+++ b/item-counter/src/infra/clientProps.ts
@@ -9,6 +9,7 @@ import type {
 	AzureLocalConnectionConfig,
 	ITelemetryBaseLogger,
 } from "@fluidframework/azure-client";
+
 import {
 	AzureFunctionTokenProvider,
 	azureUser,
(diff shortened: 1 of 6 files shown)


item-counter/src/infra/clientProps.ts
```diff
@@ -23,12 +24,12 @@ if (!useAzure) {
...
 	type: "remote",
-	tenantId: process.env.AZURE_TENANT_ID!,
+	tenantId: process.env.AZURE_TENANT_ID ?? "",
 	tokenProvider: new AzureFunctionTokenProvider(
-		process.env.AZURE_FUNCTION_TOKEN_PROVIDER_URL!,
+		process.env.AZURE_FUNCTION_TOKEN_PROVIDER_URL ?? "",
 		azureUser,
 	),
-	endpoint: process.env.AZURE_ORDERER!,
+	endpoint: process.env.AZURE_ORDERER ?? "",
```
tylerbutler, line 32: Arguably the null assertion is more correct - I could disable the rule inline here if preferred.

steffenloesch: I'd prefer to avoid the rule disabling here.

steffenloesch (approved):

PR #1859, opened 2026-04-13
build: adopt @fluidframework/eslint-config-fluid v9.0.0

tylerbutler:
## Summary

- Adopts `@fluidframework/eslint-config-fluid@^9.0.0` across all three projects (brainstorm, item-counter, item-counter-spe)
- Replaces manual FlatCompat-based `eslint.config.mjs` files with native flat config `eslint.config.mts` importing the shared `recommended` config
- Removes 6 redundant devDependencies per project now transitive through the shared config (`@eslint/eslintrc`, `@eslint/js`, `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `eslint-plugin-react`, `globals`)
- Adds rule overrides appropriate for example apps (import-x/no-internal-modules, explicit return types, filename-case, etc.)

**Depends on:** merge the per-package code fix PRs first, then rebase this branch.

## Test plan

- [ ] `npm run lint` passes with 0 errors in all three projects
- [ ] `npx syncpack lint` passes at root


steffenloesch:
This is now implemented via #1975. I think we can abandon this PR.

PR #1977, opened 2026-08-07
fix(ci): install examples through Azure Artifacts

markfields:
## Description

Route all three FluidExamples `npm ci` installations through the authenticated `publicPackages` Azure Artifacts feed instead of connecting directly to `registry.npmjs.org`.

The pipeline creates a temporary npm user configuration, authenticates it with `npmAuthenticate@0`, and passes it explicitly to the Brainstorm, Item Counter, and Item Counter SPE installs. Bash tasks are used because `Npm@1` replaces the npm configuration for custom commands.

Cold-cache pipeline validation and the CI readiness check were not run, as requested.

## Reviewer Guidance

Please verify that the public project build identity has access to the `publicPackages` feed and that the feed retains npmjs as an upstream source.

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -26,6 +26,11 @@ pr:
   - main
 variables:
   skipComponentGovernanceDetection: true
+  # Azure Artifacts feed used for every "npm ci" install below, instead of the public npm registry.
+  publicPackagesRegistryUrl: https://pkgs.dev.azure.com/fluidframework/public/_packaging/publicPackages/npm/registry/
+  # Temporary, authenticated .npmrc used by every "npm ci" install below so all package installs
+  # resolve through the Azure Artifacts feed instead of the public npm registry.
+  userNpmrcPath: $(Agent.TempDirectory)/example-packages.npmrc
 resources:
   repositories:
     - repository: 1ESPipelineTemplates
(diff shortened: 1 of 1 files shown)


build.yml
```diff
@@ -64,14 +70,41 @@ extends:
                 inputs:
                   version: 22.x
 
+              - task: Bash@3
+                displayName: Create .npmrc for Azure Artifacts feed
+                env:
+                  NPM_CONFIG_USERCONFIG: $(userNpmrcPath)
+                  NPM_REGISTRY: $(ado-feeds-primary-registry)
```
markfields, line 77: These env variables may not be necessary, instead we could use the pipeline vars directly

build.yml
```diff
@@ -64,14 +70,41 @@ extends:
...
+                    echo "always-auth=true" >> "$NPM_CONFIG_USERCONFIG"
+
+              - task: npmAuthenticate@0
+                displayName: Authenticate to Azure Artifacts feed
+                inputs:
+                  workingFile: $(userNpmrcPath)
+
               # Install, build, lint, and test each package
               - ${{ each parameter in parameters }}:
-                  - task: Npm@1
```
markfields, line 69: Using Bash instead of Npm task type is needed here since Copilot is trying to be DRY and avoid writing 3 npmrc, one for each of the example packages.

If it's preferable to do it the other way I'm happy to switch it.

markfields:
Pipeline's not expected to pass until at least Wednesday morning, due to usage of quarantined packages.

Abe27342 (approved):

